PR #2, opened 2023-03-19
Update README.md

bjornoleh:
I don’t suppose we have ambitions of making iAPS CE approved? I suggest removing anything that implies so.


Jon-b-m:
Fixed. Your PR is out of sync

PR #4, opened 2023-03-20
view Sensitivity when picking a TempTarget

mountrcg:
just as support to the discussion in https://discord.com/channels/1020905149037813862/1062353496160088215/1087327796080156772

I could not get it to work (yet) but wanted to demonstrate willingness to work in PR's instead of just discuss. So this is work and thinking in progress rather than a result to be merged.


Jon-b-m:
You should start from new dev. I just had a big push. These small bits of code you posted isn’t adding any function. And why did you remove translations? If you want to use GitHub iAPS as a playground it’s better to PR to alpha branch.

mountrcg:
back to discuss

PR #5, opened 2023-03-21
fix rounding to 0 decimals

mountrcg:
will give full mg/dL targets and mmol/L will just convert fine, also setting in mmol/L works as intended in single decimal digits


Jon-b-m:
Thanks. Did you test it also with mmol/L?

mountrcg:
mmh, the status line shows the TT in single digits for mmol/L but history shows 2 digits - even for the ones I entered in mmol/L (the middle one at 18:52)
Simulator Screen Shot - Dev iPhone 13mini - 2023-03-21 at 16 54 58

mountrcg:
could that be the conversion in History tab and the setting there to 2 digits?

PR #6, opened 2023-03-21
Create Sports TT with variable Sensitivity and TT

mountrcg:
this is long way from finished. Only have the input & calculation done and shown in the Slider View. There is still work to be done:

- the preference for HalfBasalTarget needs to be set **<- I have no idea how to do this yet**
- the NaN when the calculation hasn't been run yet should be replaced with current HBT **<- done**
- sensible limits to the slider to set **<- done**
    - if TT is <100 the desired sensitivity needs to be above 100%
    - if TT is >100 the desired sensitivity needs to be lelow 100%
- doesn't work for mmol/L  **<- done**

Unfortunatly over my non-programmers head. If someone could help please?


mountrcg:
Simulator Screen Shot - Dev iPhone 13mini - 2023-03-21 at 11 50 26

mountrcg:
so I got this workin for mmol/L and mg/dL targets. Calculations are all fine. Missing now is

1. apply diffeent limits depending wether target is a low target or high target
  - low target: Override is between 100% and autosens_max*100
  - high target: Override is between 15% and 100%
2. apply the change of Half Basal Target to Preferences

mountrcg:
making progress. The Slider now only lets you adapt Sensitivity according to Preferences Set and TT chosen. For High TT you can only reduce Insulin Ratio not increase - and vice versa.
If no settings are made in Prefs that let you change sensitivity with choosen target the Slider is hidden.

Simulator Screen Shot - Dev iPhone 13mini - 2023-03-23 at 18 08 39

mountrcg:
still missing the part how to set Preferences in that View

mountrcg:
@Jon-b-m could you have a look at it? I do not think it is really necessary to automatically change the HBT setting from the TT view. As there are more parameters that are needed to make it work. All the toggles to change Sens with TT could be off and so on.

So it is either change all of the in this view:
* HBT via the slider
* autosens_max via entry
* highTTraisesSens, lowTTlowersSens, ExcerciseMode via toggles

or non at all and users have to go to the preference settings.

mountrcg:
also I noticed that there is a method to use mmolL or mgdL in preferences. I marked in FreeAPS/Sources/Modules/PreferencesEditor/PreferencesEditorProvider.swift in comments which preference settings could need this migration as well. That would mean mmol guys could enter a hbt in mmolL as well as this SMB_above_BG setting from 0.71
What do you think @Jon-b-m ?

Jon-b-m:
Current Code (not this but in main and dev branch of iAPS) already changes "highTTraisesSens, lowTTlowersSens, ExcerciseMod" automatically. The purpose of your PR was to make it easier, no? Suggesting user to change 4 settings aren't making it user friendly or less confusing. I think your intention with PR was to change all settings automatically, HBT also, no? I suggest simplifying the UI and UX, making it understandable for everyone and easier to use.

mountrcg:
OK will check in main code. May have missed the setting of prefs there - which would help a lot as I can deduct how to change hbt from that. Thanks Jon.
 I think we should make it visual, that prefs were changed on the fly using any slider.

mountrcg:
> Current Code (not this but in main, dev and alpha branch of iAPS) already changes "highTTraisesSens, lowTTlowersSens, ExcerciseMod" automatically. 

May I ask where? I noticed you changed the default in Preferences to "true" for "highTTraisesSens" & "lowTTlowersSens" in b667d9801cf2f0e2e55412ecdc0837e8042f71f9 (*), but I can't seem to find any coding that changes the prefs when using the slider.

*) I find it dangorous to change both parameters in the default setting. Peaple should actively set these parameters to true, so that they know what is going on. Also the popup says "... defaults to false...".

Jon-b-m:
That’s why the default is not changed. Just temporarily while using the TT created with the slider.

25 mars 2023 kl. 14:54 skrev Robert ***@***.***>:



Current Code (not this but in main, dev and alpha branch of iAPS) already changes "highTTraisesSens, lowTTlowersSens, ExcerciseMod" automatically.

May I ask where? I noticed you changed the default in Preferences to "true" for "highTTraisesSens" & "lowTTlowersSens" in b667d98<https://github.com/Artificial-Pancreas/iAPS/commit/b667d9801cf2f0e2e55412ecdc0837e8042f71f9> (*), but I can't seem to find any coding that changes the prefs when using the slider.

*) I find it dangorous to change both parameters in the default setting. Peaple should actively set these parameters to true, so that they know what is going on. Also the popup says "... defaults to false...".

mountrcg:
oh, are you changing it in oref with the percentage enabled var. Thats nice.
However if so, no wonder I cant see it - please upload oref Jon - local or not 'it's not a big issue if you go cmd-a, cmd-c in your current file and cmd-a, cmd-v in 9786f6384ab96819348e4ef27f1662010d80ce3d ! takes about 20s, but I am fucked without it.

Jon-b-m:
PR it too incomplete and broken currently. I can't do anything with it.

PR #9, opened 2023-03-24
Eventual BG with 1 decimal

bjornoleh:
The eventual BG value is a highly uncertain estimate, hence it makes little sense to present this number with two decimal places. 

"numberFormatter" was replaced with a new "eventualBGFormatter", since the former is used in many places, and I did not want to check the implications of changing this formatter for all.


Jon-b-m:
It will be considered. We're planning a big UX/UI makeover, but will not disclose or release until complete.

PR #10, opened 2023-03-24
Declutter legend by hiding the obvious: green dots are of course CGM BG values

bjornoleh:
-A minor change to clear up unnecessary info


Jon-b-m:
For new users, BG is too important too to be implied. I will not remove this now.

PR #11, opened 2023-03-24
TempTargets: Better visibility and consistent colouring of temp targets

bjornoleh:
The colour .loopGreen is used for the temp target icon, and should also be used with temp targets displayed in the BG chart.
The default rendering is too dark and hard to see.


bjornoleh:
Ok, but the main point is that for consistency, the rendering of TTs should be in the same colour as the TT icon (green). Originally it’s in blue, same colour as basals. This colour changed to be less readable for TTs with my previous change for icicles, but it’s still blue. Here is the original rendering before my changes. It has better readability, but in blue:

927F46FC-8D57-47F9-A472-7BA4650FD2BE

Jon-b-m:
It will be considered. We're planning a big UX/UI makeover, but will not disclose or release until complete.

PR #15, opened 2023-03-27
Replace FAX with iAPS in build workflow, instructions and Fastfile 

bjornoleh:
- Change file name from build_FAX.yml to build_iAPS.yml
- Edit run-name: Build iAPS ${{ github.ref_name }} 
--> displays branch name for Actions runs

### DIFF
--- a/.github/workflows/build_iAPS.yml
+++ b/.github/workflows/build_iAPS.yml
@@ -1,5 +1,5 @@
-name: 4. Build FAX
-run-name: Build FAX
+name: 4. Build iAPS
+run-name: Build iAPS (${{ github.ref_name }})
 on:
   workflow_dispatch:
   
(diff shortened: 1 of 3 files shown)


Jon-b-m:
Only superficial ?

bjornoleh:
> Only superficial ?

Yes, these changes are only cosmetic, to rename from FAX to iAPS. It’s tested and works as expected. Also with current dev, the only known issue is a visual bug for selecting icons. You can still select, but the option are not shown. 
122D4768-F699-4299-A5A2-581A5D183846


The only thing to note is that the workflow file for building was renamed. You will only see it in the Actions tab in GitHub if it is in your default branch. So if someone wants to see this with dev, they CGM an change default to dev. Or wait until it’s in main. This is just how Actions work, you only see the workflows that are present in the default branch on your fork.

Issue #20, opened 2023-04-05
Experimental TempTarget not always working / Rounding Issues 


mountrcg:
I have tested the new TempTarget with adjustable HalfBasalTarget functionality extensively. It often falls back to the HBT from preferences instead the one from the oref2_variables:
Desired set ratio 50% with TT 180, which should set HBT at 180.

- loop was calculated
- Variables are saved properly in oref2_variables, is enabled true and hbt at 180
- but autosens reports 33% instead of 50%, 33% would be calculated if HBT set at 140, which is the current value in preferences

Dev Branch, latest commit / TT set in Experimental / Popup after 1st Loop / fielecontent oref2_variables
<img src="https://user-images.githubusercontent.com/539276/229986545-7be85e02-6969-4c33-9f50-e6b1a6cb3334.png"  width=250 /> <img src="https://user-images.githubusercontent.com/539276/229987304-bd3f0eb4-162c-4196-a013-b6ab98c37890.png"  width=250 /> <img src="https://user-images.githubusercontent.com/539276/229987614-0b8061e9-2eb5-4185-b6d7-24e1b21bdb58.png"  width=250 /> <img src="https://user-images.githubusercontent.com/539276/229987779-eed2f7e0-042c-4ddf-afd1-183d4dec8ac5.png"  width=250 /> <img src="https://user-images.githubusercontent.com/539276/229987950-9e314a33-e61d-4259-b5cf-cc73c2bc8dc6.png"  width=250 />

I cannot ascertain what the problem is, as I also observe proper setting of the Insulin ratio with the TempTarget SensitivityRatio being calculated with the correct HBT. However as @Jon-b-m  chooses not to publish Oref Source code there is nothing I can do at the moment to help fix this.

It's easiest tested with Ratios set to 50% as TempTarget will be the same as Halb Basal Exercise Target setting.

However to more easily analyze this issue I will put forward 3 PR's. Non of those PR's fixes the issue, though.
1.) show the TT based Sensitivity adjustment in InfoPanel besides the TempTarget -> #21 

2.) fix rounding issues that outputs calculation results of mg/dL TempTargets with 1 digit and also with 2 digits with both measurement units in the history if one switches units in Preferences -> #22 
[...]

PR #21, opened 2023-04-05
Fix TT rounding issue described in #20

mountrcg:
Fixed calculated mg/dL experimental TT from 1 decimals to 0
Fixed converted mg/dL TT in History from 2 decimals to 0.


Jon-b-m:
You don't need to have rounding 4 times. lowTarget = highTarget. 
And this PR doesn't round old presets. Preferably rounding should only be used in one place. ComputeTarget() is probably the most appropriate place for the rounding.

PR #22, opened 2023-04-05
Show insulin Ratio from Exercise Mode in Info Panel

mountrcg:


Jon-b-m:
This is not a pretty solution and not a how to handle CoreData in a SwiftUI View

PR #23, opened 2023-04-05
Simplified UI to enter TempTarget with desired insulin Ratio

mountrcg:
Step 1:
I think when doing Sports one always knows which TT to set, so no need for a slider and switching back and forth between 2 sliders.
Step 2:
After entering TT setting one adjusts the slider to set the desired Sensitivity in the same sense like an Override, sliding the desired percentage of insulin that shall be used compared to regular sensitivity.


Jon-b-m:
This is not the way forward in UX we’ve planned for

PR #24, opened 2023-04-05
Display meal presets only when FPU is enabled

bjornoleh:
The Meal Presets seems useful for complex meal entries containing combinations of carbs, protein and fat. For plain carb entries when FPU is not enabled, the presets seem redundant. This PR simplifies the Carbs screen by hiding "Meal Presets" and "Save as Preset" unless Fat and Protein (FPU) is enabled

Another benefit with this change is that the "Save and continue" button is available without scrolling or hiding of the keyboard.

### DIFF
--- a/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
+++ b/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
@@ -78,8 +78,10 @@ extension AddCarbs {
                         }
                     }
                 }
-                Section {
-                    mealPresets
+                if state.useFPU {
+                    Section {
+                        mealPresets
+                    }
                 }
 
                 Section {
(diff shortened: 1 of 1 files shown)


avouspierre:
I tested the PR with success. I agree the use without FPU should be easily and the preset options seems not really useful in the simple case of Carbs. OK for me.

Jon-b-m:
I agree in part. I will test moving the presets below the "Save and continue" button. I added the presets to this simple view by requests from other users.

Jon-b-m:
Pushed new test to dev branch.

bjornoleh:
Another alternative is to have a toggle on the carbs screen to hide or display the FPU entry + presets. In my PR this is currently done by the enable/disable FPU setting. This settings toggle could be moved or duplicated in the carbs screen (in sync with the toggle in settings). 

This would make it easier to use FPU sometimes without going to settings, and still keep the carb entry screen clean when not needed

Jon-b-m:
If you want that, then I suggest doing a new PR. The goal of your first PR I now consider completed.

PR #26, opened 2023-04-14
Add NFC tag reading capability to TestFlight app from Fastfile

bjornoleh:
This should remove the need to manually set the NFC capability in Apple Developer (https://github.com/Artificial-Pancreas/iAPS/blob/main/fastlane/testflight.md#add-nfc-tag-reading-to-freeeaps-app-id). 

There is little info on this online, so I am unsure if this works correctly (I already have my apps set up). But I have tested that it does not break anything.

It would be good if someone who have not built iAPS with Xcode->TestFlight nor GitHub Actions->TestFlight could verify if the NFC capability will be set by this change.

Also made some other minor changes to the workflows, to avoid pinning the Xcode version, but use the latest stable version available on macos-12 instead.

The changes have been briefly tested and should be safe to merge immediately.

### DIFF
--- a/.github/workflows/add_identifiers.yml
+++ b/.github/workflows/add_identifiers.yml
@@ -12,9 +12,9 @@ jobs:
     needs: secrets
     runs-on: macos-12
     steps:
-      # Uncomment to manually select latest Xcode if needed
-      #- name: Select Latest Xcode
-      #  run: "sudo xcode-select --switch /Applications/Xcode_13.0.app/Contents/Developer"
+      # Uncomment to manually select Xcode version if needed
+      #- name: Select Xcode version
+      #  run: "sudo xcode-select --switch /Applications/Xcode_14.1.app/Contents/Developer"
       
       # Checks-out the repo
       - name: Checkout Repo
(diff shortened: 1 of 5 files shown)


Jon-b-m:
why macos-12? macos-13 is latest sand stable.

bjornoleh:
> why macos-12? macos-13 is latest sand stable.
> Xcode 14.3 requires macos-13...

Yes, but unfortunately macos-13 is not yet available to GitHub Actions. Hopefully it will become available soon:

https://github.com/actions/runner-images/issues/6426#issuecomment-1479974889

Jon-b-m:
Ah. There are some Xcode 14 required changes in iAPS unfortunately.

bjornoleh:
> Ah. There are some Xcode 14 required changes in iAPS unfortunately.

Anything else than the icon selection?

Jon-b-m:
I’m using the new single image feature for icons introduced with Xcode 14. This is likely why the icons aren’t working when building with Xcode 13. Other than this I’m not 10 certain what/if other parts require Xcode 14. Will have to check…

PR #44, opened 2023-05-10
Move useFPU toggle from FPUConfigRootView to AddCarbsRootView

bjornoleh:
An attempt to make FPU more accessible without "always" being displayed.

The toggle is here more of a clickable button. Currently it is displayed as "Use FPU" with blue text, and when active, it is highlighted with black text on blue background. I am open to suggestions for using different colours, or a different text.

The "Use FPU" text is not localized (I don't know how to do that yet).

### DIFF
--- a/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
+++ b/FreeAPS/Sources/Modules/AddCarbs/View/AddCarbsRootView.swift
@@ -102,7 +102,10 @@ extension AddCarbs {
                 }
             }
             .onAppear(perform: configureView)
-            .navigationBarItems(leading: Button("Close", action: state.hideModal))
+            .navigationBarItems(
+                leading: Button("Close", action: state.hideModal),
+                trailing: Toggle("FPU", isOn: $state.useFPU) // .foregroundColor(.red)
+            )
         }
 
         var presetPopover: some View {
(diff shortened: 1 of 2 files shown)


bjornoleh:
IMG_3884

Jon-b-m:
The placement is interesting, but many users don't know what FPU means...

bjornoleh:
Ok, I think I like that, testing "Fat and Protein" now. This is also in the sequence that @GerhardFuehrer suggested, and the same as in "FPU". So perhaps the list of entries should be changed like this, although that is unrelated to this PR

1. Carbs
1. Fat
1. Protein

Jon-b-m:
Yes. You can change this here also if you want to...

Jon-b-m:
You can remove the .foregroundColor from the DecimalTextField. The colour was already declared higher up.

bjornoleh:
Ok. 

Looks pretty good, I think. 

image

Jon-b-m:
Yes, this works I think.

Jon-b-m:
Just some cleaning up in FPUConfigStateModel needed.

bjornoleh:
Ok, perhaps you can do that, and maybe also add the localisation? Haven't figured out that yet.

Jon-b-m:
> Ok, perhaps you can do that, and maybe also add the localisation? Haven't figured out that yet. 

I can't edit your PR directly (I don't think I can). You removed the toggle. So now just remove the variable and stop subscribing to it in fpuConfigStateModel. Then please test If still working. 

The localization I can add later. In another PR.

Issue #141, opened 2023-07-11
Override Profile Target field value converts into `mg/dl` after creating a new Profile


bashjs:
**Actual Behavior**
After creating a new Profile, the `Override Profile Target` field converts into `mg/dl` units.

**Expected Behavior**
After creating a new Profile, the `Override Profile Target` value should keep user-defined glucose units.

<details>
<summary>Open/Hide</summary>

|Screenshot
|-
|![IMG_4001](https://github.com/Artificial-Pancreas/iAPS/assets/17122466/783cb932-9920-4b32-8bef-56efd4fdb5fe)

</etails>

bashjs:
I'll fix it till the end of the week.

bashjs:
Fixed in https://github.com/Artificial-Pancreas/iAPS/pull/153

PR #153, opened 2023-08-03
ISSUE-141 | Fix Target Glucose value after saving a new Profile when units set to mmol/L

bashjs:
ISSUE URL: https://github.com/Artificial-Pancreas/iAPS/issues/141

Currently, after saving a new Profile the code performs an unnecessary conversion of the `target` state variable into `mg/dL` if a user is using `mmol/L` before assigning it to `saveOverride.target`.
The fix removes the conversion and assigns the converted value directly to `saveOverride.target`.

|Before|After
|-|-
|<video src="https://github.com/Artificial-Pancreas/iAPS/assets/17122466/50d46466-6ab1-4d23-b84a-1d9502118925" />|<video src="https://github.com/Artificial-Pancreas/iAPS/assets/17122466/8d4a41ab-87e6-4b81-999a-78dc52c968d5" />


Jon-b-m:
Thanks!

PR #173, opened 2023-08-14
Allow for configuring Preferences setting *defaults* via ConfigOverride.xcconfig

scrappy:
This code builds and was tested via Simulator, with expected results confirmed.

    Move Preferences.swift defaults to a Preferences.plist file.

    This allows for changing defaults via ConfigOverride.xcconfig using
    PlistBuddy as part of the Build Phase Run Script.

    Defaults in Preferences.plist are the same as they were, with the
    ability to change them being made via ConfigOverride.xcconfig as
    simple as adding:

    PS_enableUAM = YES

    I've only setup two so far, as a sample.

    This makes it easier to create a 'group' build of the app that has
    specific changes to the Preference Settings without having to make
    any modifications to the actual code base.

And:

    Fix Run Script so that it works consistently on XCode 15, by
    moving it before the Link stage, and having PlistBuddy apply to
    ${SRCROOT} *instead* of ${BUILT_PRODUCTS_DIR}


scrappy:
Am closing this ... after talking to Jon about it, he pointed to a *much* cleaner way to do what I want do, without having to modify *any* code ...

For documentation purposes, there is a directory:

`${SRCROOT}/FreeAPS/Resources/json/defaults`

That you can put "initialization" json files ... they only apply when you start the app, but every file/directory you can find in the Files app for iAPS exists in there, so you can modify, build and deploy and when you run app for first time ( ie. new phone, never configured ), it will use those settings to initialize the structure ...

The .json file has to be complete ... you can't just pick which variables change from default and leave the rest ... "as the preferences.json file is on your phone, so it has to be in that directory"

If you wipe out and re-clone, just save everything under FreeAPS/Resources/json/defaults and then recopy it back there after ...

PR #177, opened 2023-08-15
Fixes for Resources/json/defaults/*

scrappy:
Fix defaults in Preferences.swift to match pop ups
Sync Models/* -> json/defaults


Jon-b-m:
You're missing one variable in freeaps_settings defaults. Check out Models/FreeAPSSettings in iAPS to see all FreeAPS Settings (and Preferences). 

Missing variable: 
*var onlyAutotuneBasals: Bool = false*

scrappy:
It wasn't missing when I submitted it :)

Will add and update PR though ...

PR #205, opened 2023-09-05
Add a 'Push Settings' button to the Nightscount Settings screen that

scrappy:
does a force upload of preferences / statistics

Moved the check for "later then 20:00" out of statistics(), and gave a parameter to statistics() to "force" the run, overriding the "once per day" check ...

tested on my phone, can press the button repeatedly and it is updating, no crashes, and even tested going into preferences, changing my maxIOB and force the upload, works like like a charm.

the only thing I don't like, and you may have a clean idea on how to do, is that when you press 'Push Settings', it doesn't *appear* to do anything, even though I know it does.

but it does do as expected / desired.


Jon-b-m:
> does a force upload of preferences / statistics
> 
> Moved the check for "later then 20:00" out of statistics(), and gave a parameter to statistics() to "force" the run, overriding the "once per day" check ...
> 
> tested on my phone, can press the button repeatedly and it is updating, no crashes, and even tested going into preferences, changing my maxIOB and force the upload, works like like a charm.
> 
> the only thing I don't like, and you may have a clean idea on how to do, is that when you press 'Push Settings', it doesn't _appear_ to do anything, even though I know it does.
> 
> but it does do as expected / desired.

There are several ways. You can add an alert "Uploaded" to confirm the upload to NS. And you can add a tactile sensation (vibration) when pushing the button:
*let impactHeavy = UIImpactFeedbackGenerator(style: .heavy)
 impactHeavy.impactOccurred()*

Jon-b-m:
The confirmation alert "Data Uploaded" now will be displayed whether uploaded to not. All that it takes is to push the button and have whatever string (working or not) entered as URL.

Jon-b-m:
Thanks!

PR #239, opened 2023-10-09
Add Upload Preferences to Settings -> Nightscout

scrappy:
Add section to Nightscout for Manual Uploads, starting with just uploadPreferences

Simplify the whole pushSettings code from before ... uploading statistics out of schedule isn't critical as uploading preferences.


Jon-b-m:
Why not just add preferences to the already automatically uploaded profiles? 
There already is a button for upload profile manually to Nightscout in debug options. And if we add preferences to the existing upload this button will of course also upload the preferences, when you want to.

PR #247, opened 2023-10-11
Upload settings to Nightscout

scrappy:
Add functions to push FreeAPS Settings to Nightscout

to test, I added a link under Debug Options, but this should be added under the 'On Change' code that was added via upload_preferences branch ...


Jon-b-m:
Auto-upload of profiles, settings and preferences already implemented in dev branch. And the button function has been updated.

PR #266, opened 2023-10-21
Config.xcconfig: Include ../../ConfigOverride.xcconfig

bjornoleh:
Includes ConfigOverride.xcconfig from two directories up as the default configuration. This is helpful when swithcing branches, and is the default fro Loop and other apps too.


bjornoleh:
Ready to merge. Should not affect anyone negatively, but we should give some information about checking if people have ConfigOverride.xcconfig in the directory two levels up from their iAPS clone.

PR #236, opened 2023-10-06
UI/UX improvements: Reworked History and Manual Glucose Dialog, added Non-Pump Insulin Entry, Various other Additions

dnzxy:
## Contents
This PR reworks the entire treatments and glucose history. It contains the following changes
* ~~adds a new flag for non-pump insulin (also uploaded to Nightscout)
image~~
see PR #267 
* adds a new dialog to add non-pump insulin → moved functionality to treatments history(!)
* reworks the manual glucose entry dialog
* adds a toggleable feature to display treatments history in 2 or 3 tabs
   * Three tabs with no coloured dots (= the **default**):
      1. insulin treatments and pump events
      2. carb and FPU entries
      3. glucose
   * Two tabs, i.e. current display
     1. all treatments (insulin, pump events, carbs, FPUs)
     2. glucose
* removes `Add bolus without actually bolusing` functionality from the bolus entry dialog
* changes information display for entire history (alignment of text)
* adds toggle to `Statistics and Home View` settings to switch between 2-tabbed or 3-tabbed history
*  and manual glucose entries
* adds delete dialog to glucose entries, i.e., streamlining it to how treatments are deleted
* removes ⓧ icon from treatments and adds standard iOS swipe-to-delete behaviour
* adds proper descriptive message when deleting FPU entries with the sum of carb equivalents that are to be deleted
* ~~adds title to carb entry view, as it was missing~~
see PR #269 
* ~~streamlines all buttons (centered alignment, same size) for all coming-from-home-view dialogs/views~~
see PR #269 
* ~~fixed display of (temp) basal bars in dark mode on home screen, i.e. has proper blue-ish color now~~ 
see PR #268 
image

## Changes: Before and After
| **Old 2 Tab Layout** | &nbsp; | **New 2 Tab Layout** | &nbsp; |
|--------|--------|--------|--------|
| image | image | image | image |

| **New 3 Tab Layout** | &nbsp; | &nbsp; |
|--------|--------|--------|
| image | image | image |

| **Old Glucose Entry** | **New Glucose Entry** |
|--------|--------|
| image | image |

| **Old "external" Insulin entry in bolus entry** | &nbsp; | &nbsp; |
|--------|--------|--------|
| image | image | image |

| **New Non-Pump Insulin Dialog** | &nbsp; | &nbsp; |
|--------|--------|--------|
| image | image | image | 

| **Old FPU delete alert** | **New FPU delete alert** |
|--------|--------|
| image | image | 

[...]

PR #268, opened 2023-10-21
Fixed chart display of (temp)basal in darkmode

dnzxy:
This PR fixes the display of (temp) basal in the chart for darkmode in iAPS. 
It holds additions contained in PR #236, so that that PR can get smaller. 

**Quoting from PR #236:**
> fixed display of (temp) basal bars in dark mode on home screen, i.e. has proper blue-ish color now
image


Jon-b-m:
And how does it look in light mode?

dnzxy:
Exactly like it did before. 

Light mode
image

Dark mode 
image

It’s only a fix for the dark mode; as there we didn’t have a blue color specified and so it was a white background with opacity 0.4 which looked light gray.

Jon-b-m:
OK. Good 👍

PR #269, opened 2023-10-21
Streamline navigation views

dnzxy:
* Add title to carb entry view, as it was missing
* Streamlines all buttons (centered alignment, same size) for all coming-from-home-view dialogs/views
* Fix typo in variable name
* Decrease opacity to make background color better visible (more contrast for better accesibility)


Jon-b-m:
Do you have screenshots of before and after opacity? In light and dark mode?

Jon-b-m:
"Add Carbs" title was removed a long time ago due to the added Fat and Protein entries, to be able to see and enter all entries without scrolling.

Jon-b-m:
The centered alignment of button text is OK. This is approved 👍

dnzxy:
> Do you have screenshots of before and after opacity? In light and dark mode?

Sure thing. It's not *that* noticeable, but it adds a little more contrast, and I think this is important to make the app a bit more accessible. See the below table for a comparison (left old, right new), please. 

| 20% opacity (current `main` and `dev`) | 30% opacity (proposed by this PR) |
|--------|--------|
| image | image |
| image | image |

Side note: the basal bars in the chart still are "off" here, as PR #268 fixes that.

> "Add Carbs" title was removed a long time ago due to the added Fat and Protein entries, to be able to see and enter all entries without scrolling.

It's inconsistent in that *all* other views have a title and this view does not. I think we need to find a way to make it consistent, or else it looks "off". How about doing it like so for all the views that are accessible from the home view (i.e. carb entry, temp target, bolus entry, statistics, settings, history)? Or we could even make it so for the entire app, which just saves screen space for users.
image

> The centered alignment of button text is OK. This is approved 👍

Great 😄

Jon-b-m:
OK, the new opacity looks OK. I'll take a look at the carbs view later when I sit in front of a working Mac...

dnzxy:
I you would like I can provide a table here with before and after for all the mentioned views? I know you are having Mac issues, so happy to help out here and provide more examples 🤝

Jon-b-m:
The Proposed header title style makes more sense here I guess... When coming to bolus screen this is what you'll see also, since the view is normally auto-scrolled to make room from keyboard..

dnzxy:
Hi @Jon-b-m, I updated the PR with the changed header style to all views that are accessible from home screen (add carb, add temp target, add bolus, statistics, settings). 

After @LiroyvH's comment under PR #271 I also changed the layout for the ℹ️ in the add bolus screen a bit to save some screen space and get rid of the empty row. Hope you like it. 

Views look like this now
| Changed views | &nbsp; | &nbsp; |
|--------|--------|--------|
| image | image | image |
| image | image | image | 
| image | image | image |
| image | image | image |

Jon-b-m:
This look good. But please use same symbol for ℹ️ as before

Jon-b-m:
Having the blue and red next to each other is distracting

Jon-b-m:
If you revert to old symbol I'll merge it now.

dnzxy:
Reverted to old symbol and pushed 😄 

> This look good. I'm not sure why you changed to a new symbol, but I can change that later.

I changed it only because I felt that the "speech bubble"-style info icon looked a bit lost there and the filled icon looked better. I can agree on the blue looking weird. It's back to how it was now. Happy to get this merged. Once you do, I will update the other PRs with the swipe-to-delete behavior 🚀

PR #271, opened 2023-10-22
Add non-pump insulin dialog

dnzxy:
* New entry dialog to add non-pump insulin in history view
* Removes `Add bolus without actually bolusing` functionality from the bolus entry dialog
* New dialog adds ability to back-date insulin entries
* Backdating is limited to past dates up to current date

This feature was inititally proposed through PR #236. Please see that PR for user feedback (about a dozen people have tested this) and test feedback. Many users requests a button with text and icon, so I added it back (it was removed from 236).

| Non-Pump Insulin Dialog | &nbsp; |
|--------|--------|
| image | image |
| image | image |
| image | image |
| image | image |


Jon-b-m:
1. Adding insulin belongs in the insulin section, not in the pump history. We already have an insulin button for this. The current insulin view has lots of room for improvement. 
2. If you want to change the UI here you can add a top row to the list with another background and a small button fro future entries. Please see Daniel's post in Discord as an inspiration.

Jon-b-m:
Here’s some working quick drafts of how I think the data table could be designed: 
image
image
image
image
image

dnzxy:
Hi @Jon-b-m , thanks for the feedback. I'll reply one by one. 

> 1. Adding insulin belongs in the insulin section, not in the pump history. We already have an insulin button for this. The current insulin view has lots of room for improvement.

I have to strongly disagree with this notion due to various reasons. I'll list them below.
1. The current bolus entry view (you refer to it as *insulin section*) is for performing boluses from the connected pump/pod via the connected pump/pod controller, so the phone and iAPS. This means, the bolus entry view holds everything that the phone as a pump controller does there, so any insulin entry and bolus calculation should live there. 
2. Non-pump insulin is not administered via the phone and the pump, so it doesn’t really belong in the bolus entry view.
3. The current notion and action, “add without bolusing“, is a trap that can cause confusion simply by tapping the wrong button. This *could* be improved, probably, but as said before, the user action of adding non-pump insulin shouldn't live in the bolus entry view.
4. Most importantly, the main culprit users currently see with the entry, is that we have no ability to *back date entries* and I think it’s really bad UI to display a datepicker in the bolus screen when you are administering insulin from the pump at time of entry. It can cause a lot of confusion to suddenly be able to change the timestamp of insulin that is *now* being dosed from the pump/pod.
5.  Looking at use cases, you mostly record non-pump insulin when you have to have an injection by pen (often after pump fail), so you will already be in the history section to delete faulty / leaky entries that were given by pump/pod. It feels natural to also be able to add non-pump insulin here.
6. The bolus entry view is for making an *action from the connected pump*. History is for viewing and deleting treatments and glucose. Adding non-pump insulin there as a *historical record* is in line with how manual BGs are added. In both cases, this should be possible for both the current time or a past time, as one will not always be able to log it right away when taking the bolus or measuring the finger stick BG.
7. The current implementation/suggestion of a new non-pump insulin entry dialog that lives in History has now been tested by **a dozen iAPS users** and they have been testing it **for up to a month** now. Among them are @MikePlante1, @AndreasStokholm, @bjornoleh, @mountrcg, @t1dude, @tqb43, @flyingpie101 to just name a few. The entire feedback from testers has been *in favor* of the new entry dialog and how it helps them in their daily routine. 
8. To my personal surprise, @dsnallfot, who has also been a tester and who you have mentioned has provided some inspirational stuff on Discord, even adapted most of the features, that were initially suggested by PR #236, *including* the non-pump insulin entry dialog, to tweak many parts of the iAPS UI/UX. He has received praise from school nurses, other caregivers and grandparents (so elderly, probably not-so-techy users), for how it eases their use of the iAPS and makes caring for @dsnallfot 's diabetic child much easier.

Let's please listen to our user base and to the feedback they have provided us with. 

> 2. If you want to change the UI here you can add a top row to the list with  a small button for future entries. Please see Daniel's post in Discord as an inspiration and my comments about new glucose view in other PR.

I really like the button placement as a first "header" row, as you had already suggested for PR #272 and will add this accordingly 🤓 

The future entry filtering was dropped from the initial PR #236 , because it is giving issues with deleting entries (index offset) and because we had agreed, that FPU entries will be gone once we can refactor iAPS to make use of a fat-and-protein model and record *meals* rather than carbohydrates and carb-equivalents for fat and protein. Do you still think it's a valid addition? If so, I can look into it again 👍

LiroyvH:
@dnzxy @Jon-b-m 
Seems this is in a bit of a stalemate, if I understand correctly the debate is a bit about where this should be situated and both ideas have their own arguments/pros & cons. 

I'm curious if this idea is acceptable as hybrid solution:
1.) Rename the "Enact Bolus" title to "Add Insulin".
2.) Similar to the "History" window, we make it tabbed and have two tabs: "iAPS" and "Manual" (or whatever.). Then:
- In the "iAPS" tab, selected by default, you see the Bolus Recommendation and can choose "Enact bolus". We remove the "add insulin without actually bolusing", as anything administered outside of iAPS should be added in the Manual-tab.
- In the "Manual" tab, we implement dnzxy's proposal and view of how the manual bolus entries should look and work (I think they look great.). We call it "Log manual bolus" rather than "Log non-pump insulin". (Why? Because it could still be pump insulin if it concerns Medtronic users whom are not using Event History who gave a bolus through the pump itself.)

Done. Everything is now:
1.) In the same place, you have a one-stop shop by pressing the insulin button at the bottom of the screen.
2.) A cleaner interface with less clutter.
3.) As a bonus, we should move the (i)nformation button in the iAPS window to live right next to "insulin recommendation" instead of using up a full row of its own. (Ok I admit I'm just sneaking this one in here because I don't like that waste of space. :P) 

Would that be workable/acceptable? Then we have everything in one place, BUT we still have it clearly separated from each other; each with their own benefits as outlined above/as requested by the users to be clear like this.

Jon-b-m:
Yes, I like this idea.

dnzxy:
I can work with this idea, too. If you agree it’s something you will accept, I‘ll start on it.

Jon-b-m:
I can agree about the danger of inadvertently adding external instead of new directly to pump. With this in mind I will accept change of this button not the history list. 

This would work for me : 
image

This way it's (more) coherent with the current dev glucose list.

dnzxy:
Okay, I can adjust my PR accordingly and update the code. So we will keep with the dialog. I don’t think reshuffling the layout like this makes the info easily graspable. I think the layout proposed in #273 makes sense in this regard. Give me the chance to rework this please and I will ping you with the changes ready.

dnzxy:
Please also keep in mind that this is step 1 of 2. Step 2 is still to provide users with a 3 tab layout. I will just bring that in when we are done with PR #271 , #272 , #273 and others.

dnzxy:
Hi @Jon-b-m discussed changes have been pushed. 

| Header | Header |
|--------|--------|
| image | image |
| image | image |

This PR still uses the "old" way of deleting things (not swipe-to-delete, but ⓧ icon). So we should probably get PR #273 merged first. Then I can update this one and PR #272 and then those to can get in 🚀

Jon-b-m:
I don't like that row you suggested.  it looks sort of half baked compared to my suggestion.

Jon-b-m:
I have swipe to delete and the hide and show future entries already implemented in above suggestion. So you'll need to "compete" with this.

dnzxy:
Swipe to delete is part of PR #273. You have asked that I split the big PR up in as small parts as possible. This only PR only adds the new dialog and now the header row. The other fixes deletion.

The row is the same to yours for about 98% of it. I used other icons (was not able to find that booklet icon you used for „Display Future“ in SFSymbols) and I use the same font size and (+) icon for the elements as are currently in dev for the glucose tab. It’s just consistent. 

If we merge #273, I will pull in dev to this PR, update it, then it will hold swipe on delete.

Jon-b-m:
> The row is the same to yours for about 98% of it. I used other icons (was not able to find that booklet icon you used for „Display Future“ in SFSymbols) and I use the same font size and (+) icon for the elements as are currently in dev for the glucose tab. It’s just consistent.

it's all in the details. It just doesn't look good for a number of reasons. I used the Image(systemName: "book.pages") and for the insulin I used symbol "syringe". I used caption for text, to not distract from the import part, the data table entries. With text in .secondary style.

dnzxy:
> > The row is the same to yours for about 98% of it. I used other icons (was not able to find that booklet icon you used for „Display Future“ in SFSymbols) and I use the same font size and (+) icon for the elements as are currently in dev for the glucose tab. It’s just consistent.
> 
> it's all in the details. It just doesn't look good for a number of reasons. I used the Image(systemName: "book.pages") and for the insulin I used symbol "syringe". I used caption for text, to not distract from the import part, the data table entries, as it's implemented in glucose table in dev. With text in .secondary style, same as dev.

Did you see in the screenshots, that I am actually using two different calendar icons, one with a small ⊖ for "hide" and the other with a small ⊕ for "show"?

Jon-b-m:
Just Add and the syringe look much better and cleaner

Jon-b-m:
Are you using latest Xcode?

Jon-b-m:
You need to update your symbols app

dnzxy:
> Just Add and the syringe look much better and cleaner

Added ✅ although I cannot verify.

dnzxy:
@Jon-b-m I updated to latest macOS Sonoma and the icon still did not show. I then realized that the "issue" lies with iOS 17 vs. iOS 16 and earlier. I am always building to a iOS 16.4 iPhone 14 Pro (to verify backward compatibility of the things we add) and it seems like, that the new icon `book.pages` is simply not available pre iOS 17. 

So honestly, we cannot use it. Please let me know, what icon to change it for. I suggest using the logic that I initially had added, with the calendars with varying (-) and (+) symbols. Those icons are `calendar.badge.minus` and `calendar.badge.plus`. 

| iOS 17.0.1 build | iOS 16.4 |
|--------|--------|
| image | image |

Jon-b-m:
Sorry. In that case use your current calendar option.

dnzxy:
> Sorry. In that case use your current calendar option.

Updated 👍

Jon-b-m:
Text, Labels, Section etc don't need LocalizedString wrapper.

Jon-b-m:
change when to display warning message to:
let amountWarningCondition = state.nonPumpInsulinAmount > state.maxBolus
so that we always see why the button gets disabled when too high

PR #272, opened 2023-10-22
Rework glucose dialog

dnzxy:
* Refactor manual glucose entry from popup to sheet-based dialog
* Adds delete dialog to glucose entries, i.e., streamlining it to how treatments are deleted
* Small refactoring for DateTableStateModel: pass glucose object instead of index on deletion
* Change "Add" button to proper button with text and icon above list

This dialog uses the same design/style and UX as the proposed non-pump insulin dialog of PR #271. It uses a sheet-based layout and adds a button above the list below the `Picker` element.

Side note: A big **improvement** over popovers is that iOS users are *used to* these sheet views from a navigation perspective and that they bring a better UX than popovers imho, e.g. a user can swipe-away-to-close these sheets.

This feature was inititally proposed through PR #236. Please see that PR for user feedback (about a dozen people have tested this) and test feedback. Many users requests a button with text and icon, so I added it back (it was removed from 236).

| New Glucose Dialog | &nbsp; | &nbsp; |
|--------|--------|--------|
| Image | image | image |
| image | image | image |


Jon-b-m:
I have 2 comments.
1. I suggest having the button on first row, instead of adding a new element to this Form. This enables us to also add a similar row to the list with treatments, as has been suggested elsewhere with a show future entries button. 
image. Above is just a quick minute draft, and I think it look even better if first row is using another background, like greyish, like the one Daniel in Discord was displaying.
2. The delete glucose doesn't need an alert. The Apple UX with dragging a red row the entire way to the left is enough as .

dnzxy:
Good morning @Jon-b-m , thanks for the feedback. 

> 1. I suggest having the button on first row, instead of adding a new element to this Form. This enables us to also add a similar row to the list with treatments, as has been suggested elsewhere with a show future entries button.

I like this and will add this. 

> 2. The delete glucose doesn't need an alert. The Apple UX with dragging a red row the entire way to the left is enough as is.

This is inconsistent behaviour throughout the different delete actions in history. Given that swiping (without authentication for glucose) can also be done by accident, I think we should do it here as well. It's just consistent, that iAPS asks for some kind of user confirmation before the user can delete any entry from the app.

> 3. I don't like the refactoring of delete glucose, as you're now using more code to perform same function. More clean and stringent and Apple-ish before.

As this is one PR in a row of PRs, the entire deletion will actually end up living in a single function. You can compare the deletion from PR #236 where it was already done like that. I fully agree, it's not nice like it is now, but it will end up living in a very dynamic and nice function once we are done.

Jon-b-m:
Regarding the delete we already have swipe to delete. It's really clear as is. I pushed my suggestion to dev to make it easier for you. In fact probably better for you to close this particular PR, to avoid double work as it's already implemented in dev. The pop-up for glucose I can remove myself.

dnzxy:
> Regarding the delete we already have swipe to delete. It's really clear as is. I pushed my suggestion to dev to make it easier for you. In fact probably better for you to close this particular PR, to avoid double work as it's already implemented in dev. The pop-up for glucose I can remove myself.

Okay, I can pull in dev and fix this quickly so you don’t have to do it. Whichever you prefer.

dnzxy:
@Jon-b-m the discussed changes are implemented now. Please see latest commit `48a4862` 🚀 

## Reworked glucose list
| Dark mode | Light mode |
|--------|--------|
| image | image |

Jon-b-m:
The updated glucose design is already in dev. Please updated your PR and add just the new glucose form.

dnzxy:
If you take a look at my commit, I pulled in latest dev already but changed the colors and the icon to be consistent with PR #271 and added back the unit to specifc rows. 

Many users had suggested they need button texts with any (+) button, so it stays „Glucose (+)“ here and the units are back with the glucose values. 

It’s current dev, just updated 😊😊

Jon-b-m:
Please don't make changes to newly committed dev commits in this PR, as this makes it hard for me to merge. 

1. It makes no sense to have a button coloured grey (you don't see this anywhere in iAPS or in any other app). 
2. It makes no sense to show unit for every row (you normally never see this in a proper data table). Here's an example for from Apple (very similar top my example):
image

Conclusion: please keep the dev changes!

dnzxy:
Thanks for hinting me at the button color, that was my mistake. I thought I had changed that, oversight on my part.

It makes sense to show it, as we need a text-label for a button. Users have demanded that.
For comparison, this is xDrip4iOS, it shows unit labels per row. In my personal opinion, this makes sense. 
IMG_0215248D30C6-1

Jon-b-m:
xdrip would look cleaner with a unit header instead, but this is not xdrip.

dnzxy:
Okay. I reverted it with one last change I hope you can accept: I made it so, that the unit text is also part of the button, so the finger of the user has a larger surface area to tap in order to open the dialog. Looks like this:

| Dark mode | Light mode |
|--------|--------|
| image | image | 
| image | image |

Is it okay like so? This essentially adds:
* New dialog for glucose entry
* Button extended to include text with ⊕ icon

Jon-b-m:
Again it's already in dev 🙂. you don't need to change this dev UI. Also why are you displaying unit in blue? Used dev as base fro your PR, please. Just remove the pop-up in this PR.

dnzxy:
> Again it's already in dev 🙂. you don't need to change this dev UI. Also why are you displaying unit in blue? Used dev as base fro your PR, please. Just remove the pop-up in this PR.

Dev **is** the base for this PR. I don't really understand the issue, sorry. I've pulled it in various times now.
image

This PR builds upon dev, as all PR we push to dev do. It slightly changes a few things:
* The popup is removed.
* The unit is now blue so that *clickable surface area* for the button is bigger than just a tiny ⊕ symbol. For the users like me with clumsy fingers.
* The deletion is just changed slightly to make use of the item `glucose` instead of the `offset` index, because that will be needed once I start working on the 3 tab layout thing. If it really bothers you that this is in here, I can remove it now and add it back later; it will be needed for the other stuff.

Jon-b-m:
Sorry, but I would like for you to JUST remove the poop-up in this PR. Don't change anything else, please. Because then I can merge it instantly.

Jon-b-m:
Almost there! But you sneaked  in a few other changes not needed:

Delete  the @Published var manualGlucoseDate = Date() you added. Not needed.
Remove the debug about Health. It's in the wrong place. Doesn't belong here.
The  // Reset amount to 0 for next entry.  manualGlucose = 0  is also not needed. here. It's looks heavy-handed

Jon-b-m:
More moments;

Remove the 
.disabled(state.manualGlucose < limitLow || state.manualGlucose > limitHigh)

And instead add a 
.disabled(state.manualGlucose <= 0)

PR #273, opened 2023-10-22
Rework history layout and delete alerts

dnzxy:
* Changes information display for entire history (alignment of text)
* Removes ⓧ icon from treatments and adds standard iOS swipe-to-delete behaviour
* Adds proper descriptive message when deleting FPU entries with the sum of carb equivalents that are to be deleted

This feature was inititally proposed through PR https://github.com/Artificial-Pancreas/iAPS/pull/236. Please see that PR for user feedback (about a dozen people have tested this) and test feedback.

| OLD VIEW | dark & light mode | NEW VIEW | dark & light mode |
|--------|--------|--------|--------|
| image | image | image | image |
| image | image | image | image |
| image | image | image | image |


dnzxy:
@Jon-b-m Can we have a look at this one? If this one gets in, I can pull the then-updated `dev` into the other PRs, update them, and we can get them going too  🚀  The swipe-to-delete behaviour is content of this PR, as I split up the other PR into as small as possible parts, so this needs to get in first, then PR #271 (insulin dialog) and PR #272 (glucose dialog) can follow 😄

Jon-b-m:
Can we remove the order of HStack elements in this PR and just focus on the delete and alert message, please. 

I'm not sure about the order of elements yet. I think this part needs a more substantial change, but should be made in another Pr. 

Regarding the alert message if you look at the mother alerts in iAPS there are usually a \n added before every message to make the alert more readable.

dnzxy:
Sure, will remove text ordering and adapt alert text styling.

dnzxy:
I will add the history text layout/ordering in a PR of its own – it has been discussed and been approved over the course of a month my many community members. I'll push this to a new PR.

Jon-b-m:
1. This alert function has been deprecated. Please use the new 

alert(_:isPresented:presenting:actions:message:)

Instead. 

2. Please don't force unwrap, unless you 100 % know you can. In this case you don't need to. 
3. Is the use of fpuFormatter when filtering really necessary here? I think it's better to filter by fpuID here instead, as all FPUs belonging to same entry is identified with the same ID.

PR #283, opened 2023-10-27
Prevent Nightscout collisions from occurring when a transmitter ID is recycled

LiroyvH:
Generate unique _id value for G5, G6 and G7 to prevent Mongo collisions in Nightscout when Transmitter ID's are re-used (eg: recycling of TxID's by Dexcom OR users with reset enabled transmitters (eg: Anubis)).

Note: this does not fix the discussed issue of re-uploads generating 500 errors for users whom are already affected by issues caused by ID collisions, it only fixes the root cause that kickstarts the issue and ensures it won't happen in the future.

### DIFF
--- a/FreeAPS/Sources/APS/CGM/DexcomSourceG5.swift
+++ b/FreeAPS/Sources/APS/CGM/DexcomSourceG5.swift
@@ -148,7 +148,7 @@ extension DexcomSourceG5: CGMManagerDelegate {
                 let quantity = newGlucoseSample.quantity
                 let value = Int(quantity.doubleValue(for: .milligramsPerDeciliter))
                 return BloodGlucose(
-                    _id: newGlucoseSample.syncIdentifier,
+                    _id: UUID().uuidString,
                     sgv: value,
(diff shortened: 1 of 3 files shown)


Jon-b-m:
You don't need to change the G7, as the G7 Manager is already generating uniq IDs, using the generateSyncIdentifier() function.

LiroyvH:
@Jon-b-m 
True, but I figured it'd be nice to have everything working the same way - streamline it. (Plus I'd think sending the transmitter ID to Nightscout should be removed altogether, but eh.) If you want I can move the G7 changes out, but I think it'd be nice if all 3 of them work the same.

Jon-b-m:
Yeah, frankly I don't really understand why the transmuter iD is sent to NS... I would rather keep it as is until I understand why this was added to the G7 manager, please.

LiroyvH:
@Jon-b-m Me neither. I'd prefer if the Transmitter ID was never sent to Nightscout under any circumstance. No need to store it there. I had to re-fork the repo as something went wrong when trying to revert the G7 changes. New pull request has been opened. Thanks! :)

PR #286, opened 2023-10-27
Implement alternative Bolus calculator

polscm32:
### Features
- adds a new bolus calculator that calculates insulin based on carb ratio, iob, isf, target and glucose trend of the last 15 minutes and then applies a customizable factor
- adds a setting called "Bolus Calculator" with a toggle to be able to switch between the iAPS default bolus calculator and the new one and to customize the override factor
- in the new settings section also adds a toggle to override calculations with another customizable factor, e.g. for fatty meals like pizza -> if this toggle is on, the user can see a checkbox in the bolus calculator and tick it depending on the meal to apply the factor
- removes the code for the toolbar from the DecimalTextField class which solves the error UIViewAlertForUnsatisfiableConstraints

### Bolus calculator in dark mode

calc_dark

calc_dark_fatty

### Bolus calculator in light mode

calc_light

calc_light_fatty

### Popup that shows calculations in dark mode...

popup_dark

popup_dark_fatty

### ...and in light mode

popup_light

popup_light_fatty

### New setting "Bolus Calculator" in dark mode...

settings_dark

### ...and in light mode

settings_light


Jon-b-m:
The button for external insulin will be moved to pump history today or tomorrow, so you'll soon need to update you PR slightly (removing this button).

 The bolus info button has already been moved to the same HStack as "Recommended bolus" so you might want to consider moving yours too, to make it more coherent with the new dev. Otherwise it's looks pretty good 👍

Jon-b-m:
Please remove the + button. This will no longer be needed tomorrow in dev.

Jon-b-m:
Otherwise it looks cool. I will test this very soon

polscm32:
Ok, I've removed the plus button and the code that belongs to it 👍

Jon-b-m:
This needs some refactoring, but it's easier for me to do this later than having to explain it here in comments. 

The factor might cause some confusion, as we already have another factor in preferences for bolus percentage. I think I will solve this by only displaying the bolus percentage settings when **not** using your bolus calculator....

polscm32:
Sounds good with your idea for the bolus percentage setting for me.

Please let me know if there are other things I can do.

Jon-b-m:
Testing it now. It doesn't work for me

CR is always 0 and COB is Nan

polscm32:
Did you wait a loop cycle? for me its working fine....

Jon-b-m:
It doesn't retrieve the current CR

polscm32:
The carb ratio in popup shows always 0?

Jon-b-m:
Yes. I merged it with dev in a testing branch.

polscm32:
It worked on my phone and in simulator and @dnzxy also had no problems with carb ratio not showing up. Please correct me if I'm wrong.

polscm32:
I will see if the latest commits might broke something...

dnzxy:
I have not tested your branch in full – only looked at it when we were debugging that faulty log around the keyboard topbar. That was also some time ago, so there's a chance that more recent additions / changes by yourself might have broken something. 

I can test this tomorrow or Monday, not today, sorry.

polscm32:
No problem. But do you know if your carb ratio was retrieved in the version tested by you?

dnzxy:
> No problem. But do you know if your carb ratio was retrieved in the version tested by you?

I looks like it did, see these screenshots. I just build the version that I cloned back when we debugged.

| Bolus Calc | Commit `d75cddef` of `bolus_calc_test` |
|--------|--------|
| image | image |

polscm32:
Thanks! Since then I only updated to latest dev and did some minor changes that shouldn't affect retrieving CR.

Now installed latest version from today and it is still working for me...(previously was on the version when I started the PR, did not install latest one because currently theres no option to add external insulin which is essentially for me)

Jon-b-m:
I fixed it. I refactored the get current CR with less code

Jon-b-m:
image

polscm32:
> Yes. I merged it with dev in a testing branch.

Where can I find the test branch? would also like to test the change

Jon-b-m:
i tested anew to build from dev. Now it works..I think I merged it into wrong testing branch.

One thing I find weird and which needs a change is the carb amount displayed in your bolus calc. Now when go via add carbs, add let's say you add 100 g carbs and tap continue, your bolus calc will now show 0 carbs. 

if you want your bolus calc to be a shortcut for adding carbs (which is what I think you're currently using it as, you need to also fill in added carbs when coming from carb view. Or, if you don't intend to use it to bypass the carb view, remove the carbs amount entry here. 


Another thing which would be cool is if you come from carb view with added fat entries, let's say > 20 % of total energy (cal) your checkbox for fatty carbs will be automatically ticked.

polscm32:
Yes the carb entry field is one thing I wanted to discuss here. Right now you can just enter carbs and bolus for them (without saving them).

I don't use it as a bypass for the carb view, but more or less to just look how big the bolus would be for a certain carb amount. So my suggestion would be to remove the carb field for now, let more people test this and then based on their feedback maybe implement some kind of shortcut for adding carbs like you suggested (and maybe also fpus) or just displaying the amount entered in the carbs entry view.

polscm32:
And I agree, it would be cool if the checkbox would be automatically ticked based on the meal. 
Would you like to also see it in this PR or maybe in a following one?

Jon-b-m:
We can add that later.

Jon-b-m:
If you remove the carb amount I can merge it and fix the preferences

polscm32:
Ok I pulled latest dev and removed the carb entry field now.

PR #287, opened 2023-10-29
Add Swipe-To-Delete Behaviour to History Treatments

dnzxy:
This PR removes the ⓧ buttons and adjacent logic for History > Treatments and replaces it with the iOS swipe-to-delete functionality for SwiftUI Lists, equivalent to the History > Glucose list.

For additional information, tester feedback and discussion see PRs #273 and #236.


Jon-b-m:
Just tested. Deleting FPUs doesn't work.

Jon-b-m:
When delete works, it will delete some other entry, so quite dangerous now...

dnzxy:
Just pushed another attempt at a fix. @polsm32 is putting it to a hard test now 🤞

PR #290, opened 2023-10-30
Small fixes for new bolus calc

polscm32:
This PR adds back the "Continue without bolus" button when the bolus amount is 0. For values > 0 it will show "Enact bolus" or if the entered bolus amount is above maxBolus it will show "Max Bolus exceeded! and disables the button. In this case the popup also shows the info that maxBolus is exceeded and what the current setting/limit for maxBolus is. 

Moreover it fixes the wrong alignment of insulin doses when using "U" and removes the function for adding non pump insulin from BolusStateModel which is not longer needed there.


LiroyvH:
I thought this was already implemented, I tried the new bolus calculator yesterday and it suggested 43U whereas 25U is the max. When tapping it, it tried to populate the field to 43U and it instantly showed the dead emoji; that worked as expected. :) 

It'd probably be handy if it would respect the max bolus and simply not exceed that. I think ideally, if it suggests a value higher than maxBolus: it will populate the bolus field with maxBolus instead (in red color?) and show a warning explaining *why* the field got capped to 25 instead of the suggestion? Otherwise you keep running in to having to manually modify the bolus field whenever you run in to a situation exceeding maxBolus, which might be perceived as annoying.

polscm32:
Yeah I actually implemented the function that amount will be equal to maxBolus if amount > maxBolus in my previous commits but rolled it back now. Problem I had with this was that you right now don't really have the ability to confirm the amount that you put in the DecimalTextField. This way it doesn't get updated to maxBolus if you exceed it but just stays at the value you entered. So you have to open the popup for example and only then it gets updated. 

So if you have a maxBolus of 6 and enter 5 and you want to bolus, pump will give 5, but in the decimal textfield remains the 6... 

Tried several things to optimize this, for example via alert or via applying an optional limit to the DecimalTextFieldClass but after all I found this to be easier. It is currently nearly the same way loop handles it. When you exceed maxBolus, you get an alert (here the button text changes) and can not proceed.

But more thoughts on this are greatly appreciated!

Issue #280, opened 2023-10-25
[Bug] Dexcom BG readings fail to upload to Nightscout due to non-unique _id field


LiroyvH:
My environment: Dexcom G6 (Anubis), iPhone 13 Mini with iOS 17.1, iAPS-dev, Nightscout self-hosted on private server. 

_- Problem:_
I, seemingly out of the blue, started encountering issues with iAPS no longer being able to send glucose values ("entries") to Nightscout. Treatments and manual BG entries work fine, openAPS info works fine, but sending glucose values results, according to iAPS's logfiles, in a 500 Internal Server Error. Other apps, such as xDrip4iOS, would have no problem at all uploading glucose values; so a problem with Nightscout itself appeared to be ruled out. Several users reported this issue with iAPS, though only a very small fraction.
<br>

_- Digging & Potential solution:_
Upon further inspection (a packet capture (tcpdump) + analyzing it), it turns out the 500 error by Nightscout included a note containing the exact error triggering the Error 500. It appears that MongoDB threw an error "11000", which means: "Trying to write a value/index that is not unique". The only field in the "entries" collection that must be unique is the _id field, meaning somehow the exact same _id already existed in the database and thus Mongo rejected to insert our entry. 

iAPS generates the ID for Dexcom G6 as follows: https://github.com/Artificial-Pancreas/iAPS/blob/a0daab8e9118baecc2ad2173be495d1c2aa6b033/FreeAPS/Sources/APS/CGM/DexcomSourceG6.swift#L156 which results in `<Dexcom TransmitterID> + <a 6-digit number>` being set for the _id, eg: "8ABCDE 123456". Apparently this 6-digit number isn't always unique, at least not in some scenarios (see below), to the TransmitterID. What I did was replace that line (156) to read: `_id: UUID().uuidString,` instead to ensure it's random and unique.
<br>

_- (partial) Victory: Nightscout uploads of BG entries are now working again!_ 
The _id value appears to be unique again and Mongo accepts them.
[...]

PR #300, opened 2023-11-03
Fix Nightscout uploads being held back for FSL users

LiroyvH:
Fixes introduced to patch problems with non-unique ID's in Nightscout accidentally ensured that FSL1 and FSL2 users would no longer see their values (BG) uploaded to Nightscout as they do *not* get a valid UUID.

This commit streamlines the _id field with Dexcom G5, G6 and G7 and ensures it has a unique UUID in compliance with commit https://github.com/Artificial-Pancreas/iAPS/commit/7c7693c05332667efdb4a703aa4fe2cc1f979dca by @avouspierre.

This should fix the problems mentioned by @JohnKitching and @antoinekh under issue https://github.com/Artificial-Pancreas/iAPS/issues/280. 
As I do not have an FSL2, I was unable to verify/test this.


JohnKitching:
Fixed it for me thank you

PR #306, opened 2023-11-05
Alternative bolus calculator: Color coding

dsnallfot:
- Implement same warnings based on prediction as in default boluscalc
- Color coding of suggestion based on warning and insulincalculated
- This is my first PR targeting the bolus calculator. Upcoming changes will focus on color coding and formatting of the calculator popup

dark green
dark orange
dark orange warning
dark red
Light green
Light orange
light orange warning'
light red

### DIFF
--- a/FreeAPS/Sources/Modules/Bolus/View/AlternativeBolusCalcRootView.swift
+++ b/FreeAPS/Sources/Modules/Bolus/View/AlternativeBolusCalcRootView.swift
@@ -11,6 +11,7 @@ extension Bolus {
         @State private var showInfo = false
         @State private var exceededMaxBolus = false
         @State private var keepForNextWiew: Bool = false
+        @State private var displayError = false
 
         private enum Config {
             static let dividerHeight: CGFloat = 2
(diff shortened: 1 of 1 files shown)


Jon-b-m:
This bolus calculator doesn’t care about the predictions, like the eventual glucose displayed in these alerts taken from the old calculator. From reading and seeing these alert the user will likely thing these predictions are accounted for when calculating the recommended amount. 

I don’t think we should have any alerts at all. Color-coding looks nice, but any text explaining why color is changing needs to be adapted for the current calculator algo. Like what would happen if entering the recommended amount. These alert doesn’t do this (they are for the recommendations from the default calculator). 

I think removing the alerts and adding the relevant info to the pop-up instead would be better. Ideally a chart would be nice and ideally wouldn’t require any text at all.

dsnallfot:
Ok, then the issue with my suggested implementation is with the orange color. The alternative bolus calculator are only red or green since the only prediction part of it is the 15 min trend. Either there is a calculated bolus needed, in some cases restricted by a maxbolus setting, or there are no bolus needed (due to the variables in the calculation, high IOB for instance). 

The only way (currently) to capture the "orange" situations when extra consideration is required before bolusing, is to look at the existing predictions. This could ideally be visualized in a chart as you say. But we already have Eventual glucose and the different prediction as  visible part of the app in the loop circle popup, and that should be familiar to the users? 

Maybe a simplification of the text in the warning could be a way forward (not getting inte actual numbers of the predictions, just stating that there is a prediction needed to take note of before enacting the bolus? Otherwise I dint see how the orange suggestion could be explained...

Jon-b-m:
There are no predictions in this new algo. The trend is just history of 3 last glucose readings. 

The old calculator were including predictions in calculating the recommended amount. Like if you give this amount your eventual BG will not get under target.

Jon-b-m:
Therefore we can't use same alerts

dsnallfot:
I understand. I will close this PR until we have a solution on how to bring predictions into the new calculation.

dsnallfot:
@Jon-b-m  Maybe one "simple" way to do it would be to show an orange color and warning every time the alt boluscalculator comes up with a bigger bolus number than insulinRequired or insulinRecommended?

PR #323, opened 2023-11-10
Add Timezone Offset Indicator

dnzxy:
**Update
* Changed this PR with a condition fix and a changed positioning after great feedback on Discord
* Updated images in table below to represent the current look of UI changes in PR

----

## Content

This PR adds a small indicator (~~SF Symbol `clock.fill`~~ SF Symbol `clock.badge.exclamationmark.fill` in `.symbolRenderingMode(.multicolor)` with `.warning` as primary and `.red` as secondary colors) that gets displayed ~~in the center of the Loop circle icon~~ next to the pump's or pod's reservoir status when there is an offset between the iPhone's current timezone and the programmed pump time. This usually happens when traveling between timezones or when time's are changed from DST. The icon uses the `.warning` color that is currently also being used when the reservoir gets low on insulin, for example.

## UI Change
These screenshots are taken with the Macbook's time set to GMT+1 while the pump time is on GMT-8. 

| Dark theme | Light theme |
|--------|--------|
| image | image |

### DIFF
--- a/Core_Data.xcdatamodeld/Core_Data.xcdatamodel/contents
+++ b/Core_Data.xcdatamodeld/Core_Data.xcdatamodel/contents
@@ -68,7 +68,9 @@
     <entity name="Meals" representedClassName="Meals" syncable="YES" codeGenerationType="class">
         <attribute name="carbs" optional="YES" attributeType="Double" defaultValueString="0.0" usesScalarValueType="YES"/>
         <attribute name="createdAt" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="date" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
         <attribute name="fat" optional="YES" attributeType="Double" defaultValueString="0.0" usesScalarValueType="YES"/>
+        <attribute name="fpuID" optional="YES" attributeType="String"/>
         <attribute name="id" optional="YES" attributeType="String"/>
         <attribute name="note" optional="YES" attributeType="String"/>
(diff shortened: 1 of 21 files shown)


Jon-b-m:
Only run this extra pump status command when timezone changes, instead of continuously. Set up a timezoneDidUpdate() for this, as has been done for the other variables in Home View.

dnzxy:
> Only run this extra pump status command when timezone changes, instead of continuously. Set up a timezoneDidUpdate() for this, as has been done for the other variables in Home View.

I have pushed some changes. Please let me know if this looks more like what you are looking for :)

Jon-b-m:
It's looking better.

I think it would be better to use a iOS system colour for the warning, as the system colours are keeping a high contrast (>4.5) in both light and dark mode and auto-adapts to the current background, but it's a minor thing I can change later. 

I'll test this PR later, thanks.

dnzxy:
@Jon-b-m Great! I agree about the color, I think you are right. That warning color is currently also used for the insulin reservoir drop - maybe we should also change that to a better color with less contrast. Let me know if you find anything that needs changing 😊

avouspierre:
@Jon-b-m  this code breaks the capability to enter a carb in the past. in AddCarbsStateModel : 
` let carbsToStore = [CarbsEntry(....
                id: id,
                createdAt: Date.now,...
` 

to be revert to 

`
let carbsToStore = [CarbsEntry(...createdAt: date, ...
`

Jon-b-m:
It was a mistake merging in latest dev changes in this PR by me. Let me check this...

PR #346, opened 2023-11-16
Make "Max IOB 0" warning more prominent

dnzxy:
## Description

On many occasions, there have been confused iAPS beginners around the warning `Max IOB: 0` that is displayed on the home screen. It's been one of the more frequently asked questions on FB and it was also prominently explained on one of the Loop'n'Learn YouTube iAPS Walkthrough videos. Recently, a user on Discord thought that this mean "No limit" – and was wondering why there were no SMBs given. So this still seems to be an issue!

## Content
This PR introduces a small warning icon and changes the text from orange to make this warning more prominent. It makes it more obvious that there is something "wrong", when a user has closed loop enabled and still has max IOB set to 0. It also adjusts to `.padding` to be consistent with all other elements in the top right corner of the info area, so `.padding(.trailing, 8)`.

## UI Changes
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/HomeRootView.swift
+++ b/FreeAPS/Sources/Modules/Home/View/HomeRootView.swift
@@ -310,7 +310,10 @@ extension Home {
                 }
 
                 if state.closedLoop, state.settingsManager.preferences.maxIOB == 0 {
-                    Text("Max IOB: 0").font(.callout).foregroundColor(.orange).padding(.trailing, 20)
+                    (Text(Image(systemName: "exclamationmark.triangle")) + Text(" Max IOB: 0"))
+                        .font(.callout)
+                        .foregroundColor(.red)
+                        .padding(.trailing, 8)
                 }
 
                 if let progress = state.bolusProgress {


dsnallfot:
I’m a big fan of color coding as you know 😊 Red is great to draw attention to “this is an issue you really need to act on” where yellow/orange is more of a “notice this” color by my definition.

So it makes sense to use red coloring since max iob 0 practically puts the system in limp/PDM-mode and disables most automatic correction (shut of basal and temp basal/smbs if IOB<0 still works with max IOB 0 I figure?)

However. The exclaimationmark.triangle symbol is more of a yellow/orange warning symbol as I see it. Maybe we should use the x.circle instead to match the red status? (To align/open up for potential more color coding where we make a distinction between yellow and red, with exclaimation vs x warning symbols?)

dnzxy:
If I understand you correctly, you would like to introduce the following:

| Icon | Color | Meaning |
|--------|--------|--------|
| image | Red | Error |
| image | Orange | Warning |
| image | Green | Success | 

I think for warning, a more consisting icon would be this one (however, I think the triangle with exclamation mark is a more generally recognized symbol for warning).
image

I can agree, that making it red may be too much of a "THIS IS AN ERROR" type message. This is *actually* a warning, not an error I would say. Let's see, what Jon says. Happy to change it back to orange colouring.

dsnallfot:
Yes exactly! I was thinking beyond just the max IOB 0-warning. 

If we want to use the same logic for success-warning-error consistently in other parts of the app, it's good to agree on guidelines for which symbols and colors to use for which situations (level of severity). 

I personally prefer the exclamation.triangle for warnings since the x.circle and exclamation.circle is harder to distinguish at a quick glance, and as you said, the triangle is commonly used for warnings elsewhere. But let's see what Jon says.

Jon-b-m:
The real issue here is not about the use of a  triangle or not. It's about some users not understanding the Max IOB setting. Ideally the **real** solution to this problem is introducing an onboarding for this setting (and ideally for all of the relevant settings). I would welcome this effort.

dnzxy:
Hi everyone. Is is still being considered moving this along as an intermediate improvement for "Max IOB" until a proper onboarding process is in place, or is this an unwanted proposition?

Jon-b-m:
If you want to make a real fix, please do. This is not good one, not even as an intermediate fix. I'll close this now, as there is a better intermediate fix in the pipeline already. Coming next week I think along with other improvements.

PR #370, opened 2023-11-22
Pinch to zoom on chart view

10nas:
As shown on discord:
- Smooth zooming between 2 and 24 hour screen hours scale.
- On end of zoom gesture the screen hours are rounded to neatest full hour of the final zoomed state. So this allows setting the screen hours to 2h, 3h, 4h, ...
- A nice side-effect of the changes in this PR is that pressing the preset screen hour buttons (2h, 4h, 6h, ...) no longer has any lag on the chart re-layout.

A few notes:
- Only supports iOS 17 due to requiring `MagnifyGesture`. The old `MagnificationGesture` doesn't provide the necessary data.
- Sometimes a pinch might not be properly detected due to the scroll view thinking that it should start scrolling. Generally not a big issue IMO. Perhaps try to see for yourself. Only way to make zooming and scrolling play more nicely together seems to be to use an actual UIScrollView, which I didn't want to do unless absolutely necessary (since it might introduce a different set of issues).
- I want to further expand on this feature in a separate PR by adding a "bolus in window" info text that updates as you scroll and zoom you can see the sum of boluses currently shown on screen. This will allow to easily figure out the total insulin used for a past meal when there were several different boluses delivered for that meal (e.g. a pre-bolus followed by SMBs during/after the meal). When there are many subsequent SMBs it becomes hard to keep track of the total amount that was delivered and this feature will be aimed at resolving this lack of insight.

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/Chart/MainChartView.swift
+++ b/FreeAPS/Sources/Modules/Home/View/Chart/MainChartView.swift
@@ -54,6 +54,12 @@ struct MainChartView: View {
         static let bolus = "💧"
     }
 
+    private struct ZoomState {
+        let scale: Double
+        let anchorTime: TimeInterval
+        let anchor: UnitPoint
+    }
+
     @Binding var glucose: [BloodGlucose]
     @Binding var isManual: [BloodGlucose]
     @Binding var suggestion: Suggestion?
@@ -86,20 +92,45 @@ struct MainChartView: View {
     @State private var unSmoothedGlucoseDots: [CGRect] = []
     @State private var predictionDots: [PredictionType: [CGRect]] = [:]
     @State private var bolusDots: [DotInfo] = []
-    @State private var bolusPath = Path()
     @State private var tempBasalPath = Path()
     @State private var regularBasalPath = Path()
     @State private var tempTargetsPath = Path()
     @State private var suspensionsPath = Path()
     @State private var carbsDots: [DotInfo] = []
-    @State private var carbsPath = Path()
     @State private var fpuDots: [DotInfo] = []
-    @State private var fpuPath = Path()
     @State private var glucoseYRange: GlucoseYRange = (0, 0, 0, 0)
-    @State private var offset: CGFloat = 0
     @State private var cachedMaxBasalRate: Decimal?
+    /// updated while pinch to zoom gesture is in progress
+    /// this gets automatically reset when the gesture ends or is cancelled
+    /// on cancellation the result is that the `effectiveScreenHours`  just jumps back to the pre-gesture state
+    /// on end the `screenHours` are updated, `additionalZoomState` is set temporarily to have a smooth transition to the final `screenHours` state
+    @GestureState private var zoomGestureState: ZoomState? = nil
+    /// caches the last non-nil `zoomGestureState` which allows the pinch gesture to transition smoothly to the final `screenHours` on gesture end, when `zoomGestureState` has already been reset to nil
+    @State private var lastZoomGestureState: ZoomState? = nil
 
-    private let calculationQueue = DispatchQueue(label: "MainChartView.calculationQueue")
+    /// zoom state that outlives a pinch gesture.
+    /// used to smoothly transition to the final quantized zoom scale (`screenHours`) on gesture end
+    @State private var additionalZoomState: ZoomState? = nil
+
+    /// the effective zoom state
+    private var zoomState: ZoomState? {
+        additionalZoomState ?? zoomGestureState
+    }
+
+    private var zoomScale: Double {
+        (1.0 / Double(screenHours)) * (zoomState?.scale ?? 1)
+    }
+
+    /// currently displayed screen hours. Includes currently applied scaling from in-progress pinch to zoom gesture
+    private var effectiveScreenHours: Double {
+        Double(screenHours) / (zoomState?.scale ?? 1)
+    }
+
+    private let calculationQueue = DispatchQueue(
+        label: "MainChartView.calculationQueue",
+        qos: .userInteractive,
+        attributes: .concurrent
+    )
 
     private var dateFormatter: DateFormatter {
         let formatter = DateFormatter()
@@ -164,10 +195,6 @@ struct MainChartView: View {
             .onChange(of: vSizeClass) { _ in
                 update(fullSize: geo.size)
             }
-            .onChange(of: screenHours) { _ in
-                update(fullSize: geo.size)
-                // scroll.scrollTo(Config.endID, anchor: .trailing)
-            }
             .onReceive(
                 Foundation.NotificationCenter.default
                     .publisher(for: UIDevice.orientationDidChangeNotification)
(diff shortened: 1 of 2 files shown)


dnzxy:
While the feature does look nice and smooth, bumping the requirement to use this to iOS17 is a no-go in my opinion. If we cannot support at least 16.0+, then we might as well stop pursuing this feature. Backwards compatibility is vastly important.

10nas:
> While the feature does look nice and smooth, bumping the requirement to use this to iOS17 is a no-go in my opinion. If we cannot support at least 16.0+, then we might as well stop pursuing this feature. Backwards compatibility is vastly important.

This PR doesn't change the minimum required version to iOS 17. The pinch to zoom feature is simply only made available on devices running iOS 17 while devices on iOS 16 just don't get the feature. See https://github.com/Artificial-Pancreas/iAPS/pull/370/files#diff-a1829f13c9e47fb5b022acab720cdcad4a8299f1253788ddc44ff8d08b90c2faR282

dnzxy:
@10nas I‘m aware, I looked at your code before commenting. 

What I‘m calling into question here is investing manpower and resources (yours), while I‘d expect that the majority of the user base is not running iOS17 (yet). Also, in the intermediate future the current chart will hopefully be ported to Swift Charts, so your implementation may be broken. I’ve DM‘d you regarding portability to Swift Charts, you haven’t responded.

10nas:
A lot of iOS users upgrade their OS very quickly, I'd imagine those who are tech savvy enough to use iAPS even more. But you have a point, I agree it would be better to have this feature run on iOS 16 too.

Regarding Swift Charts: From my side, this PR is ready to go as-is. So whether it is merged now or only after a change to Swift Charts is done will not make a difference for me in terms of work I have to invest at this point.

Now would be a chance to get some initial feedback though (e.g. on the interplay between pinch to zoom and the hours picker bar) so that I could incorporate those changes when also adding support for the change to Swift Charts.

mountrcg:
I will test, I have such a wild diverse unreleased iAPS code base, that this fits right in - :-)

Jon-b-m:
Thanks for PR! 

I think If we use this we don’t need the time buttons, which saves us some screen real estate, which is good. I will test your PR very soon

Jon-b-m:
And I agree about ios17. 80-90 % of iPhone users always updates to latest OS

dnzxy:
As someone who provides a lot of support and is also moderator in one of the two big DIY AID iOS FB groups, I can tell you from experience; no, they do not immediately update and, more importantly, the recommendation and what many many users abide to, is NOT to update until it has been cleared to be safe. They have been living with that „rule“ in their head since the inception of iOS Loop in 2017/18, so please understand that a large number of users is still running iOS 16.x, some still on iOS 15.x, definitely not the majority on iOS 17.x.

Furthermore, a very large number of users do not rely on or have access to the most current models of iPhones for their looping device. Access to iPhones and the burden of cost are one of the major drawbacks around the iOS ecosystem, especially if you rely on the device for such crucial, life-dependent application. Many iOS DIY users are children or teenagers with tech-savvy parents. 

An example: from a table of 10 DIY iOS users at DData, all of which are heavily involved in DIY and contribute, there were two people with iPhones >13, most had <=12, and the majority was not running iOS17. That is only an n=10 thing, and just an example but I‘m trying to make a point here. 

Thinking or expecting that providing such an important feature and improvement can or should be limited to iOS 17 is not a good thing. I would not have brought it up otherwise.

dsnallfot:
> Thanks for PR!
> 
> I think If we use this we don’t need the time buttons, which saves us some screen real estate, which is good. I will test your PR very soon

Could a compromise be to keep the time buttons for iOS<17 and replace them with this pinch to zoom-feature for iOS17?

dnzxy:
That may be an alternative. 

Just to also say it in writing: I really like the overall feature and it brings such a smooth and intuitive interaction behavior to the chart, so it’s not that I don’t want to see it. I merely am worried about the feature being limited for an imho not to be underestimated number of users that may miss out on this (if we do not / cannot provide a good alternative).

Jon-b-m:
> As someone who provides a lot of support and is also moderator in one of the two big DIY AID iOS FB groups, I can tell you from experience; no, they do not immediately update and, more importantly, the recommendation and what many many users abide to, is NOT to update until it has been cleared to be safe. They have been living with that „rule“ in their head since the inception of iOS Loop in 2017/18, so please understand that a large number of users is still running iOS 16.x, some still on iOS 15.x, definitely not the majority on iOS 17.x.
> 
> Furthermore, a very large number of users do not rely on or have access to the most current models of iPhones for their looping device. Access to iPhones and the burden of cost are one of the major drawbacks around the iOS ecosystem, especially if you rely on the device for such crucial, life-dependent application. Many iOS DIY users are children or teenagers with tech-savvy parents.
> 
> An example: from a table of 10 DIY iOS users at DData, all of which are heavily involved in DIY and contribute, there were two people with iPhones >13, most had <=12, and the majority was not running iOS17. That is only an n=10 thing, and just an example but I‘m trying to make a point here.
> 
> Thinking or expecting that providing such an important feature and improvement can or should be limited to iOS 17 is not a good thing. I would not have brought it up otherwise.
> 
> PS: I personally also do not use iOS17, nor do I own the latest iPhone mode (14/15). Of the iAPS admins on FB, that is also true for many of them. Let’s not fall victim to confirmation bias here, please.

I haven’t even tested this yet 😅, but I’m not making these numbers up.  The percentage I got from Apple own stats. They published data before iOS 17 was released of numbers of devices running iOS 16. This was 90%… 

Regarding you seeing older devices at a conference: iOS 17 is supported on iPhone XS/11, haha. And regarding ios15, this version is not supported on iAPS. 

I will test this PR later.

Jon-b-m:
> > Thanks for PR!
> > I think If we use this we don’t need the time buttons, which saves us some screen real estate, which is good. I will test your PR very soon
> 
> Could a compromise be to keep the time buttons for iOS<17 and replace them with this pinch to zoom-feature for iOS17?

Of course.

dnzxy:
> I haven’t even tested this yet 😅, but I’m not making these numbers up. The percentage I got from Apple own stats. They published data before iOS 17 was released (mars) of numbers of iPhone devices running iOS 16. This was 90%…
> 
> Regarding you seeing older devices at a conference: iOS 17 is supported on iPhone XS/11, haha. And regarding ios15, this version is not supported on iAPS.

Again, all I am trying to say is we cannot go off of official Apple Statistics but what the DIY community is doing and has been advised to do for years. That was simply: don’t just update, be careful. Plus many users don’t update for other reasons. Just because older devices can carry a newer OS, it’s often not installed for performance reasons and/or worries about making the DIY app incompatible. I’m not making this up, it’s just based on all the questions and worries we see on the FB support groups around new iOS versions. 

This is all I am trying to say here 😊

Jon-b-m:
@10nas I'm testing it now. 

Unfortunately it lags a lot (iPhone 14, iOS 17.1.1).
And often the pinch to zoom gesture isn't recognised.It now fails more often than it works, which with time would make the use case for this feature non-existent.

However I do like the refactor of the highlight() function you made. I'll keep this, thanks!

10nas:
> @10nas I'm testing it now.
> 
> Unfortunately it lags a lot (iPhone 14, iOS 17.1.1). And often the pinch to zoom gesture isn't recognised.It now fails more often than it works, which with time would make the use case for this feature non-existent.
> 
> However I do like the refactor of the highlight() function you made. I'll keep this, thanks!

Weird, I have no lag at all on 14 Pro. Do you have a lot of events on your chart (boluses, carbs, etc)?

I agree on the bad recognition of the pinch, it's not ideal. Perhaps I should wait until a refactor to Swift Charts is done and then have a go at it again. Might just have to go with UIScrollView as SwiftUIs ScrollView is just really janky.

I also recommend to use the changes I did here https://github.com/Artificial-Pancreas/iAPS/pull/370/files#diff-a1829f13c9e47fb5b022acab720cdcad4a8299f1253788ddc44ff8d08b90c2faR797 as it removes the lag when switching between the screen hours using the tab bar. Basically all my changes to that file except the pinch to zoom part. I can make a separate PR for that if you wish.

Jon-b-m:
Yes, please do. I already changed the highlightButtons

MikePlante1:
Just chiming in to say I just cloned and tested this and it worked perfectly for me and was very smooth. Then again, I'm using a 14PM, so I'm not sure how it runs on slower phones. Also, I like it how it's implemented as an addition to the buttons, not a replacement for the buttons.

mountrcg:
iPhone 13 mini (ios 17) and the pinchig is very smooth. Very noticeable how much faster it switches the scaling using the buttons.

mountrcg:
I guess iz needs some de-cluttering of y-axis at 12 hrs and bigger: e9c99a07f1c118e4f6a16b924dffed931772b46d

10nas:
Closing in favor of #394 for now

PR #379, opened 2023-11-23
Fix upload and import functions

scrappy:
Simple changes to make sure that we both import and upload offset values in the proper units.  NS uses seconds, while we use minutes.

Also, changed the importSEttings so that when it imports offset/minutes, it uses the self.offset calculated value, not the value that is pulled from NS ... that eliminates any potential for a mis-match on import


Jon-b-m:
You are now opening all the JSONs from fileStorage  twice. First in a guard where all files are opened and then individually in an if/else. Instead make a guard for each JSON, this way each JSON only opens up once and is checked individually, before continuing to the next.

Jon-b-m:
Please just use a guard let instead for each json.

Issue #381, opened 2023-11-23
[enhancement]  Replace App Explanation in Write Access to Health on Apple Watch


bashjs:
## Description

In the initial Apple Watch setup, the app asks for Write Access to Health. However, the App Explanation contains dummy text which should be replaced with some meaningful one.

## Screenshots

|iAPS Watch app|Other Watch app as an example
|-|-
|![incoming-ADFB967E-7C08-427F-8BB1-2FF403F5A3F2](https://github.com/Artificial-Pancreas/iAPS/assets/17122466/fc5c0688-f8c1-4778-bd76-c9e8c88fd96f)|![incoming-3203132F-5019-4067-BF21-74F23E847625](https://github.com/Artificial-Pancreas/iAPS/assets/17122466/325accaa-18fa-458c-93eb-ec51734d9cb3)

## Proposal

Update the App Explanation with the following text:

> APP EXPLANATION: For the iAPS app to work on the Apple Watch, please ensure that you have enabled HealthKit permissions.

Jon-b-m:
OK. Would you mind PR:ing this, please.

PR #384, opened 2023-11-24
ISSUE-381 | Replace dummy text in `APP EXPLANATION` on Apple Watch

bashjs:
#381

## Blocked by

- [ ] Add [localization support](https://developer.apple.com/forums/thread/124254)

## Changes

- Replaced `Bla bla {Record,Share,Update}` text with `For the iAPS app to work on the Apple Watch, please ensure that you have enabled HealthKit permissions.`

### DIFF
--- a/FreeAPSWatch WatchKit Extension/Info.plist
+++ b/FreeAPSWatch WatchKit Extension/Info.plist
@@ -3,11 +3,11 @@
 <plist version="1.0">
 <dict>
 	<key>NSHealthClinicalHealthRecordsShareUsageDescription</key>
-	<string>Bla bla Record Health</string>
+	<string>For the iAPS app to work on the Apple Watch, please ensure that you have enabled HealthKit permissions.</string>
 	<key>NSHealthShareUsageDescription</key>
-	<string>Bla bla Share Health</string>
+	<string>For the iAPS app to work on the Apple Watch, please ensure that you have enabled HealthKit permissions.</string>
 	<key>NSHealthUpdateUsageDescription</key>
-	<string>Bla bla Update Health</string>
+	<string>For the iAPS app to work on the Apple Watch, please ensure that you have enabled HealthKit permissions.</string>
 	<key>NSExtension</key>
 	<dict>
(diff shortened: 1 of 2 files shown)


FreeAPSWatch WatchKit Extension/Info.plist
```diff
@@ -3,11 +3,11 @@
 <plist version="1.0">
```
bashjs, line 3: What if we delete this file? Will the app pick [these translations automatically?](https://github.com/Artificial-Pancreas/iAPS/blob/dev/FreeAPS/Resources/Info.plist#L77-L80)  🤔

Jon-b-m: I'm not sure, but I remember when it was needed to add this file, hence I would suggest not to delete it. Speaking of translations, you didn't add the English source strings for the localization.

bjornoleh:
Hi, the suggested changes are not quite how these strings are meant to be. Please see some examples here:

https://github.com/Artificial-Pancreas/iAPS/blob/cb67cbef100253c965f0a30cc2582cb1260201b0/FreeAPS/Resources/Info.plist#L77

These strings are presented tu the user when the app requests permissions. The idea is that the app is transparent about why it wants you to give permissions, such as collect sensitive health data or usage of capabilities like bluetooth. For an app to pass AppStore review (not relevant for us), these things can typically be a blocker unless its done properly.

Documentation is available here:
https://developer.apple.com/documentation/bundleresources/information_property_list/nshealthupdateusagedescription

Jon-b-m:
It looks ok, but you just need to create the localizable strings also. If you create the English source file with strings  first you after that can just check all languages for the other to be created in Xcode. I can help you with this next week after Christmas.

Jon-b-m:
Please check out this PR: https://github.com/Artificial-Pancreas/iAPS/pull/670
and test it this fixes it for you?

Jon-b-m:
Looks like there are still some strings missing in the localizable.strings ?

PR #393, opened 2023-11-29
Live activity

10nas:
- Only works on iOS 16.2+ because of API availability.
- Includes setting to turn off live activity.
- Live activity is started when the app leaves the active state. Activities can be active for up to 8h. As long as app is opened at least once every 8h, live activity should work continuously.
- Activity turns stale 6m after last bg reading, at which point it will no longer display BG. This is to avoid showing old data. The lock screen and expanded dynamic island view also shows timestamp of displayed BG as further ability to verify that the displayed data is current.
- There have been mentions of using a shortcut + automation to automatically renew live activities in regular intervals. This could be useful to cover cases where the app is not re-opened every 8h. It might work with [LiveActivityStartingIntent](https://developer.apple.com/documentation/appintents/liveactivitystartingintent). But this should be left to a separate PR, if implemented at all.
- Not tested over long periods of time yet.


MikePlante1:
I love it! Great idea. Fingers crossed that it won't affect battery life too much.

Some feedback:

I think the "show live activity" toggle should probably be defaulted to off.

As others noted on Discord, delta is always in mg/dL even if mmol/L is used in the app.

Also, here's a screenshot of the Dynamic Island on my iPhone 14 Pro Max where the delta is cut off. Delta was just +2
IMG_3222

PR #397, opened 2023-11-30
Live activity fixes

### LINKED ISSUE #402: When using mmol/L live display shows BG changes in mg/dL
IMG_1915

### LINKED ISSUE #403: Live activity text is difficult to read over dark wallpaper in light mode
When you use a dark wallpaper and the system theme is dark, the text on the live activity notification is difficult to read.

A possible way to improve this might be to set the notification text color to white and the background color to black with ≈90% opacity. This should improve readability for this case while still ensuring readability over light wallpapers.

Dark mode                          |  Light mode
:-------------------------:|:-------------------------:
  |

10nas:
fix mmol live activity, live activity off by default, re-create live activity always after it ended, better readability on lock screen

Fixes #402 
Fixes #403


10nas:
Good to go @Jon-b-m

PR #407, opened 2023-12-02
Rework Alternative Bolus Calculator Popup

dnzxy:
## TL;DR
This PR reworks (and adds to) the calculations popup of the **Alternative Bolus Calculator** as of iAPS `dev`.
It introduces a completely refactored layout and adds lots of detail information around the recommended bolus calculations.

## Contents
* Changes the popover to a sheet-based view
* Introduces grid-based layout within view; not based only on VStacks and HStacks
* Introduces calculation formulas and applied math, i.e. filled in formulas with numbers, per single calculation step
* Refactors state model; removes all rounding and rounded values from state model
* Add rounding in view; performed here for formatting only
* Tested on screens down to iPhone 12 mini and up to 15 Pro Max
* Collected feedback from multiple testers (n=7) with screen sizes from iPhone SE to 15 Pro.

## Screen Recording for UI/UX Feel
Recorded on an iPhone 15 Pro simulator build. Download [here via WeTransfer](https://we.tl/t-uFf8BURm6F) due to too large file size (>10 MB).

## Screenhots

**iPhone 15 Pro with and without added meal, with and without fatty meal factor**
| Header | Header | Header |
|--------|--------|--------|
| image | image | image | 
| image | image | image |

**iPhone 12 mini with and without added meal, with and without fatty meal factor**

| Header | Header | Header |
|--------|--------|--------|
| image | image | image |
| image | image | image | 

Notes:
* We may want to rename the "factor" to something more descriptive of what it does/means. Open for discussion
* If this design is accepted, I would also rework the popup for the current "default" calculator popup in a similar style.

### DIFF
--- a/FreeAPS/Sources/Modules/Bolus/BolusStateModel.swift
+++ b/FreeAPS/Sources/Modules/Bolus/BolusStateModel.swift
@@ -45,12 +45,11 @@ extension Bolus {
         @Published var fifteenMinInsulin: Decimal = 0
         @Published var deltaBG: Decimal = 0
         @Published var targetDifferenceInsulin: Decimal = 0
+        @Published var targetDifference: Decimal = 0
         @Published var wholeCobInsulin: Decimal = 0
         @Published var iobInsulinReduction: Decimal = 0
         @Published var wholeCalc: Decimal = 0
-        @Published var roundedWholeCalc: Decimal = 0
         @Published var insulinCalculated: Decimal = 0
-        @Published var roundedInsulinCalculated: Decimal = 0
         @Published var fraction: Decimal = 0
         @Published var useCalc: Bool = false
         @Published var basal: Decimal = 0
@@ -115,7 +114,7 @@ extension Bolus {
                 conversion = 0.0555
             }
             // insulin needed for the current blood glucose
-            let targetDifference = (currentBG - target) * conversion
+            targetDifference = (currentBG - target) * conversion
             targetDifferenceInsulin = targetDifference / isf
 
             // more or less insulin because of bg trend in the last 15 minutes
@@ -140,9 +139,6 @@ extension Bolus {
                     wholeCalc = (targetDifferenceInsulin + iobInsulinReduction + wholeCobInsulin)
                 }
             }
-            // rounding
-            let wholeCalcAsDouble = Double(wholeCalc)
-            roundedWholeCalc = Decimal(round(100 * wholeCalcAsDouble) / 100)
 
             // apply custom factor at the end of the calculations
             let result = wholeCalc * fraction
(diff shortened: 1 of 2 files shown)


LiroyvH:
"We may want to rename the "factor" to something more descriptive of what it does/means. Open for discussion"

That's from the "Recommended Bolus Percentage"-setting right? Maybe just name it "Percentage" then, but it may be a bit confusing if we use factors here and percentages (recommended, easier to understand for user) elsewhere hehe. Especially as the setting for RBP lies in the Bolus Calculator settings.

Jon-b-m:
If we should change the current pop-up it needs to be to something preferable more graphical and less verbose than current. This suggestion looks quite busy, unfortunately, even compared to the current pop-up.

It would be great to have something cleaner and more illustrative to look at, something more graphical. It's too much monochromatic text now. 

The code could some boil down.  I will come back to this PR after you've done some refactoring.

dnzxy:
@Jon-b-m Please give me one or multiple examples or scribbles of what you have in mind, I genuinely have now idea, sorry. 

My idea was this: The intention with this was to break a mathematical equation into its substeps and give details per step. This was done so the user can get a full grasp of the calculation of recommended bolus with formula and all applied math. I don’t really know how you can make a mathematical equation graphical with this intention in mind, please help me understand your idea. 

As for the code: The code was completely refactored from Stack usage to Grid usage. Every single row of the grid is stored in a variable of its own. Please tell me what you would like to have boiled down. Should I add doc strings (comments) to it?

Jon-b-m:
More graphical and less verbose as in something like this: 
IMG_5609

dnzxy:
This would be a great improvement for the overall view, not the detail view, right? 

My PR only improves the detail view, so where the user wants to see exactly how the calculated recommendation was made up. Verbosity and detailed calculation is the intention here. Sorry if that got caught up somewhere - no change to the overall calculator view.

Jon-b-m:
This is  fior the pop-up, of course

dnzxy:
Okay, thanks for clarifying that. I have no idea how  that should fit in the popup, sorry. Can it be scrollable, is that okay?

flyingpie101:
you are right sorry.  I have personally tested this PR, and I was pretty happy with it as is.
Outside of the main view, I tend to disable graphics, hence why I wade in.

Jon-b-m:
I don't want to this write the PR for you, but as you don't have this clear vision/idea perhaps it would be better, instead of holding your hand every step when painting these illustration? 

Either this or you'd have to be prepared making a lot of further changes and refactoring after you added the 3 charts etc, depending on my feedback.

dnzxy:
I am asking you for your input, as you are the person wanting changes. This has nothing to do with holding hand or writing the PR for me. In a real world setting I would be working of a clear ticket, we don't have this. So I am asking you. 

Since you have such a clear idea maybe it is best you go ahead yourself here.

Issue #398, opened 2023-11-30
Possible rounding issue in glucose values


dsut4392:
I’m seeing an occasional small discrepancy between Dexcom glucose values in the main display and the History display. Main screen reported value matches what is seen in the G6 app.  Smooth Glucose Value is turned OFF in the CGM settings. 
iAPS dev 2.2.9 branch 1079fb7
Dexcom G6 mmol/L
iOS 16.7.1, iPhone 12mini

![IMG_4061](https://github.com/Artificial-Pancreas/iAPS/assets/55490253/282383e6-40f7-4e10-8cdf-34caa0d83dc0)
![IMG_4060](https://github.com/Artificial-Pancreas/iAPS/assets/55490253/84039a20-7317-4a31-80a7-48c2cbd45307)

dnzxy:
@dsut4392 exceptional find, great eye for detail! While I could not reproduce this issue in a simulator, I did find that glucose in history uses a different rounding method than basically *anywhere else* in the app. I've adjusted it to be consistent. If you could maybe check that out (PR #408 to `alpha`) and report back?

Jon-b-m:
Fixed in dev

PR #408, opened 2023-12-02
Use same rounding mode for glucose in history as everywhere else

dnzxy:
Addresses rounding offset as reported in issue #398

### DIFF
--- a/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
+++ b/FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
@@ -31,9 +31,10 @@ extension DataTable {
             formatter.numberStyle = .decimal
             formatter.maximumFractionDigits = 0
             if state.units == .mmolL {
+                formatter.minimumFractionDigits = 1
                 formatter.maximumFractionDigits = 1
-                formatter.roundingMode = .ceiling
             }
+            formatter.roundingMode = .halfUp
             return formatter
         }
 


Jon-b-m:
Example presented in my first post 🙂. 
When entering high manual glucose entries, the displayed amount is different from what you entered, due to the rounding. 

Easiest fix would be use another formatter for manual glucose entries, using the .ceiling rounding rule here. 

Do you want me to fix this for you?

dnzxy:
I read that you were saying really high values, yes, that is still very abstract. High values in mmol/L means 15-20? With decimals or without? Sorry, just trying to be able to reproduce this locally here on my end to then fix it. 

I’m still a bit confused because we are only using the `.ceiling` rounding for this specific glucose formatter in the data table view in the app. All other instances where we have distinct glucose formatters use the `.halfUp` that was introduced here. 

Does this mean that initially, when you fixed this with `.ceiling`, that it was only added in the data table but not for all other controllers where we have glucose formatiertes? So subsequently we are using wrong rounding everywhere else but here? I am also asking this in such detail, because in #398 @dsut4392 reported that the glucoses are correctly rounded and displayed (like in his Dexcom app) everywhere but the data table view. 

Happy to look at this once it is clear where and what needs fixing 😊

Jon-b-m:
I will close this PR and make the fix.  It's quicker than discussion or explaining in detail what you need to do with this PR.

dnzxy:
Clear input will produce clear output. Leaving questions unanswered doesn't.

PR #413, opened 2023-12-05
Live activity updates 2

10nas:
- Use direction string instead of system icon for trend
- Use semantic colors and improve readability
- Don't hide BG value when data is stale, strikethrough instead
- Prevent flashing live activity briefly when live activities are turned off
- Detect system setting, link to system setting when it is turned off
- Add trend to minimal view (might still get truncated when BG and direction string sum to more than 4 characters)
- Ensure live activity is always started, even when immediately stale
- Fix for iOS < 17


Jon-b-m:
Thanks for PR! Is this tested on a iOS 16 device? Looks like  you changed the minimum deployment to iOS 17?

PR #435, opened 2023-12-27
improve open loop and manual/suspend OmniPod use :

avouspierre:
- avoid to have manual temp and close loop
- display correct information in the home screen
- block open/close loop toggle when manual or suspend mode
- add correct icon wheel in apple watch


avouspierre:
I don’t know if you use sometime open loop @Jon-b-m but currently we can have a erratic behavior with open/close loop + manual or suspended omnipod option. For example, you can start a suspended pump with omnipod option and you think you are always with close loop..what’s wrong. Other example : you start a open loop and start a manual temp during 1h.. and you can close the loop again but the manual temp is in progress.. 

so not lot of changes but avoid to have strange behaviors with the option of open loop. Based on my experience of open loop user.

Jon-b-m:
I don't use open loop (I don't think any experienced looper does?), but I'm using the manual temp basal quite often while doing prolonged exercising. For this purpose it's really useful using closed loop *together* with a manual temp basal, as iAPS will continue to operate as usual in closed loop after the duration of the temp basal (the exercise). I don't want to remove the possibility for or to complicate this use case. 


Regarding your use case when having open loop and using manual basal and then closing the loop while manual basal running: in this scenario it would be better to have an alert which informs you that you're using a manual temp basal and asking if you want to continue with the manual basal or ending it when you close the loop.

Jon-b-m:
And regarding your watch commit: I would prefer if you made a separate PR for this, thank you.

Jon-b-m:
There are 0 files changed? 
Did you mean to close this PR?

Jon-b-m:
Since no response and no changes compared to dev, I'll close this PR now.

PR #436, opened 2023-12-28
Update HomeRootView.swift

nicolevanelstvandenhoek:
Improved icons and customised display of override settings (different colour icon when active)
Scherm­afbeelding 2023-12-28 om 12 54 20


Jon-b-m:
The colour when activated looks good. Could you please make a new PR with just this change? And use the same colour also for the temp targets. Thanks!

PR #458, opened 2024-01-06
Add slider button for cannula insertion and pod deactivation from Loop

itsmojo:


marionbarker:
I tested the OmniBLE portion of this modification on 6 January 2024 using both English (left to right language) and Hebrew (right to left language) using iAPS communicating with an rPi DASH simulator. The OmniKit changes are the same (per code inspection).

The sliders for both Insert Cannula and Deactivate are configured for the correct direction and work as expected for both ltr and rtl languages.

marionbarker:
The rationale is to break the flow where an inattentive user might touch a button and insert the cannula too soon, or deactivate a pod when they did not mean to.

Yes - the user is asked do they really mean it, but having to change the acceptance to a slide instead of a tap was requested for Loop. This same code is offered to iAPS with this PR.

Example screen shots are provided for a left to right language (like English) on the left and a right to left language (like Hebrew) on the right. Neither Hebrew nor Arabic have the translations needed for this to show up other than in English, but when translated, the words will appear in the correct orientation.

The sliders work in the natural orientation for each language.

iaps-sliders

Issue #457, opened 2024-01-05
Bug: Previous entered carbs could get deleted by mistake


dsnallfot:
Bug found in latest dev but have most likely been present since the edit meal from bolusview was introduced. 

**How to replicate the bug:**
When entering through bolus view and navigating from there to carbsView, and adding new carbs or cancelling (cancel button or swipe down) without adding any carbs, the most recent carb entry gets deleted/overwritten. 

I've implemented a quick and dirty fix in my personal branch to only show the meal-button in the bolus view navigationBarItem when "if fetch" is true (effectively blocking the possibility to navigate to carbs view when starting the process in bolusview, but allowing to go back and edit carbs if coming in to bolus view with already entered carbs from carbs view). 

This probably could be solved more elegantly why I'm creating this issue instead of a PR.

Jon-b-m:
Thanks for reporting this. This is indeed unexpected behaviour. Would you mind showing your fix?

dsnallfot:
Submitted a PR with my personal interim fix. As stated, probably not a perfect solution for all possible usecases.

Jon-b-m:
I'm looking at this now, but interestingly I can't replicate this bug with latest dev. 
Are you using a clean clone of dev or some modified fork? Which is your commit nr?

Jon-b-m:
Fix is in dev now. Please test it, before I close this issue as competed.

dsnallfot:
Nice! Seems to work as expected now

PR #459, opened 2024-01-06
Fix to prevent overwriting/deleting of recent carbs...

dsnallfot:
...when coming to meal view from bolus view.
(breaks possibility to start meal registration from bolusview which maybe could be relevant for users with "skip bolus screen after entering carbs" setting)

Interim quick fix for Issue #457 (could probably be solved in a more elegant way)


Jon-b-m:
Thanks, but since this brakes functionality of iAPS I will not merge this.

PR #467, opened 2024-01-08
Selecting alarm sound, + Fixed a bug 

Vasily-van-Zaam:
Fixed a bug in playing a melody in the melody selection list


Jon-b-m:
I’ve tested this during a few hours so far and seems to be working 👍. 

I would very much like you to add a separate configurations for low and high. Especially for low glucose it’s important to have a distinctive sound. And regarding high glucose, I think some users might not want a sound, and for those who do preferably a different sound than the low warning sound.

Vasily-van-Zaam:
> I’ve tested this during a few hours so far and seems to be working 👍.
> 
> I would very much like you to add separate configurations for low and high. Especially for low glucose it’s important to have a distinctive sound. And regarding high glucose, I think some users might not want a sound, and for those who do preferably a different sound than the low warning sound.

Fine. I'll try to figure out how to implement this.

Vasily-van-Zaam:
i think we need to create multiple event types.
1. low glucose level
2. high glucose levels
3. rapid drop in glucose levels
4. Rapid increase in glucose levels.
5. Closed loop error.
Because if we only have "low" or "high" it will add more questions about how it will work.

Jon-b-m:
I guess you're right. It has been requested before. If you want to do it all in one go (preferably) or dividing it up into smaller parts or to delegate some of it to me or to someone else that would also be OK.

Vasily-van-Zaam:
Fine. I'll try to find time for this. I'll slowly start studying the code to understand how to implement this.

Jon-b-m:
How is the work going for this one? Is it still in progress?

PR #470, opened 2024-01-09
Fixing data model sent over to the Garmin device

gonkowonko:
This PR resolves some issues with the data model sent to the Garmin devices.

Added: **glucoseDateInterval** - Simply the numerical date value of the date the last BG was received
Updated: **trendRaw** - Garmin code expects this to be the textual description of the trend and not special characters. 

Fixed a spelling mistake on the function eventualBGString.

IMG_6700

### DIFF
--- a/FreeAPS/Sources/Services/WatchManager/WatchManager.swift
+++ b/FreeAPS/Sources/Services/WatchManager/WatchManager.swift
@@ -60,8 +60,12 @@ final class BaseWatchManager: NSObject, WatchManager, Injectable {
             self.state.glucose = glucoseValues.glucose
             self.state.trend = glucoseValues.trend
             self.state.delta = glucoseValues.delta
-            self.state.trendRaw = readings.first?.direction ?? "↔︎"
+            self.state.trendRaw = self.convertTrendToDirectionText(trend: glucoseValues.trend)
             self.state.glucoseDate = readings.first?.date ?? .distantPast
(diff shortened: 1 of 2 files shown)


Jon-b-m:
Thanks! You're referring to "some issues", what issues are there still?

gonkowonko:
Ah sorry thats my very poor use of phrasing, for the Garmin data field for my testing it resolves all the issues. I think the biggest one was the TrendRaw going from a string to special characters. The glucoseDateInterval just makes it consistent with the loopDate property and allows the use of it (if needed) in the data field. (I use it in my customised version)

Its used in 2 different formats for the Garmin devices, as a watch face and as a datafield. The watchface didn't have any issues with what was being sent as it didn't use all the properties. Its only the data field that had issues. Below are links the the implementations so you can see its usages if needed.

**WatchFace**
https://github.com/Artificial-Pancreas/garminWatch/blob/01735f7dcef7877a60fb0e740b9d028725160bb4/iAPSGarminWatchface/source/iAPSWatchfaceView.mc#L19C1-L19C1


**DataField**
https://github.com/Artificial-Pancreas/garminWatch/blob/01735f7dcef7877a60fb0e740b9d028725160bb4/iAPSGarminDataField/source/iAPSDataFieldView.mc#L15C5-L15C5

Jon-b-m:
I changed to alpha branch instead.

PR #473, opened 2024-01-09
New release 2.3.3

Jon-b-m:
New release

* Safety fix / Info for Middleware users. Display when active middleware. Some less experienced users might forget that they are using some copied middleware code Now they will notice in red in the big pop-up. For the more advanced users this will allow for direct confirmation of an active middleware with the configurable output string in pop-up.

* Use the same "remote" for remote commands in Nightscout, as in dev branch. Instead of "freeeaps-x-remote" now use "remote". Please be careful. This is for testing. More confirmation/display of remote commands will be pushed to main later (now in dev).

* Lots and lots of new translations and new strings. Added Hungarian and Vietnamese for iAPS and all its dependencies. (Vietnamese at 100 % in Crowdin project (https://crowdin.com/project/iaps).


Config.xcconfig
```diff
@@ -1,5 +1,5 @@
 APP_DISPLAY_NAME = iAPS
-APP_VERSION = 2.3.2
+APP_VERSION = 2.3.3
```
dnzxy, line 2: This PR adds new, non-breaking changes. Subsequently it should be a minor release, not a patch-release, i.e. 2.x.0

FreeAPS/Resources/Info.plist
```diff
@@ -70,6 +72,8 @@
 	<string>Bluetooth is used to communicate with insulin pump and continuous glucose monitor devices</string>
 	<key>NSBluetoothPeripheralUsageDescription</key>
 	<string>Bluetooth is used to communicate with insulin pump and continuous glucose monitor devices</string>
+	<key>NSCalendarsFullAccessUsageDescription</key>
+	<string>To create events with BG reading values, so that they can be viewed on Apple Watch and CarPlay</string>
 	<key>NSCalendarsUsageDescription</key>
 	<string>Calendar is used to create a new glucose events.</string>
```
dnzxy, line 78: Typo here. 
„a new glucose events“

FreeAPS/Resources/javascript/prepare/determine-basal.js
dnzxy, line 1: This entire function can be rewritten like so, which is more concise and readable, and does not unnecessarily use doubly declared constants or inelegant stuff like `oref2_variables_`. Moreover it uses `const` instead of `var`, which is common practice since ECMAScript6 (2015) and makes a variable truly prone from being overridden and also block scoped.

```js
function generate(iob, currenttemp, glucose, profile, autosens = null, meal = null, microbolusAllowed = false, reservoir = null, pump_history, preferences, basalProfile, oref2_variables) {
    const clock = new Date();
    let middleware_was_used = "";

    try {
        middleware_was_used = middleware(iob, currenttemp, glucose, profile, autosens, meal, reservoir, clock, pump_history, preferences, basalProfile, oref2_variables) || "Nothing changed";
        console.log("Middleware reason: " + middleware_was_used);
    } catch (error) {
        console.log("Invalid middleware: " + error);
    }

    const glucose_status = freeaps_glucoseGetLast(glucose);

    return freeaps_determineBasal(glucose_status, currenttemp, iob, profile, autosens, meal || {}, freeaps_basalSetTemp, microbolusAllowed, reservoir || null, clock, pump_history || {}, preferences, basalProfile || {}, oref2_variables || {}, middleware_was_used);
}

```

dnzxy (changes requested):

FreeAPS/Resources/javascript/prepare/determine-basal.js
Jon-b-m: Thanks, but your comment is indicating some confusion on your end.
1. Like the comment about oref2, which hasn't changed or even being used for the added display of middleware. 
2. And let is not a constant, but a local variable. 
3. And the remark about double declared constants: I'm not even using constants? I added one, and only one variable. 

The output to logs I can agree can be simplified to:
console.log("Middleware reason: " + middleware_was_used);
but that's hardly a rewrite of the entire function, is it? I will change this in dev.

dnzxy: Hi, I was wrapping up my work day so kept the comment very short. Let me write it a bit more detailed, maybe this can clear up the confusion on your end :) 

- The function has 13 parameters of which 5 have default values in case the passed input is `null` or `undefined`.
- Subsequently, doing all this (see snippet below) is unnecessary, as it takes the function parameter, and checks for `null` or `undefined` and then declares a new variable with similar name with the passed input, hence me saying you are redeclaring.
```js
var glucose_status = freeaps_glucoseGetLast(glucose);
    var clock = new Date();

// ... other code here

    var autosens_data = null;

    if (autosens) {
        autosens_data = autosens;
    }
    
    var reservoir_data = null;
    if (reservoir) {
        reservoir_data = reservoir;
    }

    var meal_data = {};
    if (meal) {
        meal_data = meal;
    }
    
    var pumphistory = {};
    if (pump_history) {
        pumphistory = pump_history;
    }
    
    var basalprofile = {};
    if (basalProfile) {
        basalprofile = basalProfile;
    }
    
    var oref2_variables_ = {};
    if (oref2_variables) {
        oref2_variables_ = oref2_variables;
    }
```
- You are using `var` here, which is JavaScript-style that predates ES6 (<2015) and should just not be done anymore. Since you are not actually mutating data for these variables, they can be constants, hence me calling them constants, hence me suggesting to use `const` where necessary and omit all these redeclarations where unnecessary.
- So we can omit that entire block I pasted above and simply do the check for `null` or `undefined` where we are passing the data, hence in the function call. 

Rereading my suggestion and the given code, this can be simplified even more, because earlier I actually had missed, that you are already giving default values for certain parameters. 

I am not entirely sure, why you are giving `meal` a default value `null` and then redeclare it as an empty object in case of `null`, but being that as it may, the function can be boiled down a lot. 

So essentially my suggestions provide the following:

- Removes the redundant declaration of `var clock = new Date();` since clock is already declared as a parameter.
- Simplifies Middleware Handling by removing the intermediate variable `middlewareReason` and directly assigning the result of the middleware function to `middleware_was_used`.
- It uses the local OR operator (`||`) to set `middleware_was_used`
- It consolidates asignments for `autosens_data`, `reservoir_data`, `meal_data`, `pumphistory`, `basalprofile`, and `oref2_variables_`  
- It removes unnecessary conditional checks for `autosens`, `reservoir`, `meal`, `pump_history`, `basalProfile`, and `oref2_variables` since the logical OR operator handles the default values.
- It consistently uses `const` by changing the variable declarations to use `const` where appropriate for improved readability and to indicate that these variables should not be reassigned, i.e. are immutable.

Hope more clear now.

Jon-b-m: I don't understand why you're commenting on stuff not changed in this PR? The variables you're referring to are not my code, but Ivan's which has been there since forever. If you want to make new changes, make a proper PR to dev, please. So that we can review and follow what you've actually intending/suggesting.

dnzxy: I commented on it, because you extended it - and while extending it, it can _also_ be made more concise with little to no effort. No reason to get defensive over it 😊

You added a change I have suggested to dev - you can take more of my suggested changes and add them to dev. Or you can deny my proposed changes and that is also okay. 

No one is pointing fingers. I did not check the git blame who introduced what, I just saw you added some in various places in this file - it can be made a lot easier with low effort - I suggested changes 😊

Jon-b-m: Again, please make a PR with what you're actually suggesting. Because I can't see what you're thinking.

dnzxy: Okay 😅. Meanwhile, the entire suggested change is in my initial comment. You could also take a look at the code snippet there.

Config.xcconfig
Jon-b-m: I can't, since this version then will be too close to dev version, even exceeding some more recent dev versions. The dev and main release versions will get in better sync later, starting from 3.0.0, but right now it is how it is...

FreeAPS/Resources/Info.plist
Jon-b-m: Thanks. All translations are handled in Crowdin. will push to Crowdin project.

Config.xcconfig
dnzxy: You are right, good thinking. Maybe an idea: Loop has odd numbers for dev and even numbers for release. If I am not mistaking then dev is currently 2.4.x and latest main is 2.3.2. 

How about we the Loop order, and make 2.6.0 the next minor release for main, which puts dev at 2.7.0. 

All following changes to dev are then patch versions, so 2.7.1, 2.7.2, etc. and once you say it is release ready for the next minor release it is, main is bumped to 2.8.0 and dev automatically jumps to 2.9.0

I know this would skip 2.5.x, but maybe it would work? What do you think 😊

Jon-b-m: Yes, perhaps something like that. Skipping a number would indeed solve the current version confusion.

PR #475, opened 2024-01-11
Extend HealthKitManager with Logging

dnzxy:
Adds more logging and cleaner error handling to try and get to the bottom of the recent HealthKit issues. 

Thanks for @scrappy for reminding to PR this.


Jon-b-m:
It would be good with a bit more descriptive log and message. For instance instead of "Saved blood glucose" it could read "Glucose save to Health Store", because glucose is also saved in glucoseStorage and CoreData in other places. Same thing with the error message, which could read "Couldn't save glucose to Health Store ...". 

Regarding the carbs message, it's not just carbs but also other treatments, like temp basals and boluses saved to Health store. It would be good to see which treatment in logs and error message. 

There are also bugs concerning deletion of Health Store currently. It would therefore be good including logging and errors of deletion of Health Store objects.

dnzxy:
Added more verbose logging. 

In the context where the "carb entry" logging was added, it is indeed storing carb entries (function signature reads `func saveIfNeeded(carbs: [CarbsEntry]) {…}`). I think talking of a "carb entry" there is okay.

Example log entry:
`Stored insulin entry in HealthKit Store!  Metadata: Optional(Dictionary.Values([1, "f81a58640a502d7c083439309510a614", 1, true, "f81a58640a502d7c083439309510a614"]))`

All deletion functions already have logging in case of an error. Would you like success logging, too?

Jon-b-m:
Ah, yes I remember now from
seeing some logs that deletion is logged and in fact I added some more descriptive logging here already

Jon-b-m:
And yes it looks better now, thanks!

PR #480, opened 2024-01-14
Refactored generate function for clarity and consistency

dnzxy:
This PR proposes the suggested changes in #473 's PR review to `FreeAPS/Resources/javascript/prepare/determine-basal.js`'s `generate` function, in particular:

* Simplified date initialization; removed redundant var clock = new Date();.
* Streamlined middleware result assignment and logging.
* Used const for immutable variables where applicable (glucose_status).
* Enhanced error logging with console.error.
* Simplified assignment of optional parameters (autosens_data, reservoir_data, meal_data, pumphistory, basalprofile, oref2_variables_).
* Maintained default parameter values for readability and consistency.

Tested in iOS 17 simulator, tested on iPhone 15 Pro running iOS 17.1.1

### DIFF
--- a/FreeAPS/Resources/javascript/prepare/determine-basal.js
+++ b/FreeAPS/Resources/javascript/prepare/determine-basal.js
@@ -1,49 +1,19 @@
-//для enact/smb-suggested.json параметры: monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json settings/autosens.json --meal monitor/meal.json --microbolus --reservoir monitor/reservoir.json
+// FOR: enact/smb-suggested.json
+// PARAMETERS: monitor/iob.json monitor/temp_basal.json monitor/glucose.json settings/profile.json settings/autosens.json --meal monitor/meal.json --microbolus --reservoir monitor/reservoir.json
 
 function generate(iob, currenttemp, glucose, profile, autosens = null, meal = null, microbolusAllowed = false, reservoir = null, clock = new Date(), pump_history, preferences, basalProfile, oref2_variables) {
-
-    var clock = new Date();
     
-    var middleware_was_used = "";
+    let middleware_was_used = "";
+
     try {
         var middlewareReason = middleware(iob, currenttemp, glucose, profile, autosens, meal, reservoir, clock, pump_history, preferences, basalProfile, oref2_variables);
-        middleware_was_used = (middlewareReason || "Nothing changed");
+        middleware_was_used = middlewareReason || "Nothing changed";
         console.log("Middleware reason: " + middleware_was_used);
     } catch (error) {
-        console.log("Invalid middleware: " + error);
-    };
-
-    var glucose_status = freeaps_glucoseGetLast(glucose);
-    var autosens_data = null;
-
-    if (autosens) {
-        autosens_data = autosens;
-    }
-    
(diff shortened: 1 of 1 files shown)


Jon-b-m:
Reducing the code is always nice. 

However despite your verbose commit comment I don't understand why exactly you chose to remove several of the variable checks for nil, but still kept some? 

As it's a bit hard to test and since current "/prepare/determine-basal.js" doesn't contain any bug or has ever caused any issues as is, would you please care to explain?

dnzxy:
It’s unnessecary code to it like it’s currently done. Can be written shorter with the same outcome. That’s what this PR proposes. Variables that already have a default null value don’t need additional checking when passed to the next layer. Those that don’t will get assigned a proper null value / empty object and pass on to the next layer. 

I used what was there (logic wise) just made it shorter, more readable to developers.

dnzxy:
Check the other PR where I had commented that we don’t need to doubly assign variables for checking before passing them down the oref2.

MikePlante1:
I haven't tested it at all or anything, but from a quick glance it doesn't look like anything actually changes in this PR, it just makes it more concise and readable.

I don't see any reason to check those 2 variables for null, and then if it is null, set it to null... whereas in order to keep the old code's functionality the same for the other 4 checks you do still have to set null values to {} which is what the || does in the return statement.

Jon-b-m:
What the script does here is to check all arrays and variables before relaying these to oref0, as any nil value will potentially brake the loop. 

Now some of the nil checks are removed, and I just wonder why?

MikePlante1:
Looks like if `autosens` and `reservoir` are nil they're already set to null in the function's parameters, though. If you used `pump_history = {}` in the parameters it would return `null` instead of `{}`, though, which is why `pump_history||{}` is in the return statement.

I don't know why some are `null` and some are `{}` and didn't look into it, but it looks to match the original functionality.

dnzxy:
I concurs with Mike, thanks for chiming in and trying to explain the intent of this PR 😊

Not a single check was removed. The parameters that have default values are not checked again, as they will default to null, if they are not given; if null is passed they are null anyways. The parameters that are not null checked before, are then checked but in the return (that’s what the `||` does). 

I‘m not entirely sure why some parameters default / are assigned to be null, while others are assigned to be an empty object; it was done like this in the existing code, so it was done here again, I didn’t call this into question 

The code here is just more concise, clear and readable. Nothing was removed in terms of checks. The unnecessary double checks (instantiating a local variable, then checking a parameter and assigning null to the local variable) was removed - but the check is still in place, via the parameters default values and where not present in the return statement.

dnzxy:
@Jon-b-m I see that 3 days ago you labelled this with **question**. What is your question? I've tried answering your [initial inquiry](https://github.com/Artificial-Pancreas/iAPS/pull/480#issuecomment-1908702312), and there have been no follow-up questions.

JohnKitching:
> What the script does here is to check all JSONs before relaying these to oref0, as any nil will brake the loop. 
> 
> 
> 
> Now some of the nil checks are removed, and I just wonder why? 

Do you mean "brake" as in slow down the loop or "break" as in stop the loop working?

MikePlante1:
@Jon-b-m Was this accidentally closed when fixing the alpha branch, and should be re-PR'd to alpha?

PR #483, opened 2024-01-18
Refactor SMB Overrides. Use the oref enableSMB function instead, also…

Jon-b-m:
… for disabling with an override, to make code easier to overview.


MikePlante1:
Could you please post a link to the oref2 repo for the changes made here, so we can better understand the update?

Jon-b-m:
It's like the title says, It's now using the normal enbleSMB(...) oref0 function to disable SMBs also when a profile has SMB off active. Everything now at one place. I can't push to oref2 right now due to some local source control integration issues in VSCode, but I will do this later. 

Closing this PR, as commits already merged from looking at diff...

MikePlante1:
In the future, if you could please update oref2 before pushing the minified into dev or main? I find it very problematic that it's impossible to tell what oref code is actually being used in each version of iAPS main or dev, especially when those changes aren't even available anywhere online.

PR #522, opened 2024-02-10
Update defaults to match Models

scrappy:
Match both content, and ordering, of Models files for preferences and freeap_settings


Jon-b-m:
Please always PR to dev branch, from dev branch. I changed to dev for you, but due to the conflicts please make a PR based on dev instead.

PR #534, opened 2024-02-14
Revert commit 3f2eddc (Due to potential crash when cancelling ongoing…

dsnallfot:
… bolus)

### DIFF
--- a/FreeAPS/Sources/Services/HealthKit/HealthKitManager.swift
+++ b/FreeAPS/Sources/Services/HealthKit/HealthKitManager.swift
@@ -234,23 +234,24 @@ final class BaseHealthKitManager: HealthKitManager, Injectable, CarbsObserver, P
               events.isNotEmpty
         else { return }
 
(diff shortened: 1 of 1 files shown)


dsnallfot:
Also tested this revert on fresh build on test phone. Bolus cancel works. No crashes. 

note: The aborted bolus amount gets written to AH together with the next insulin entry (not immediately after cancelling)

avouspierre:
thanks @dsnallfot . The revert deleted the optimisation code avoiding to load HK twice but Don't understand why this code was OK during 2 months and now crashed the app.

dsnallfot:
I didn't figure out exactly why the bolus cancelling crashed, and I'm positive I've tried to cancel bolusing with your commit in the code without issues before. 

Maybe some under the hood change of HK in the latest iOS? However, after lots of trial and error, and as soon as I found out that this reverting this specific commit seemed to fix the crash, I made a PR by request from Jon on discord.

PR #551, opened 2024-02-25
Remove steps related to syncing of the "alive" branch

bjornoleh:
- name: Sync upstream changes
- name: New commits found
- name: No new commits
- name: Show value of 'has_new_commits

Create the "alive" branch from main, not dev

The "alive" branch is still automatically created, and "keep alive" commits will be added to the "alive" branch to avoid a stale repository.

This will avoid errors when updating the alive branch causes merge conflicts.

### DIFF
--- a/.github/workflows/build_iAPS.yml
+++ b/.github/workflows/build_iAPS.yml
@@ -69,9 +69,9 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        # Get ref for Artificial-Pancreas/iAPS:dev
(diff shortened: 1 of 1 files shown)


bjornoleh:
I have tested this successfully, both with an alive branch in place that would otherwise cause failed build due to merge conflicts, as others have reported on Discord. Then, no merges were attempted, but a keep-alive commit was inserted into the "alive" branch to avoid staleness. And also tested after deleting the alive branch, which was successfully created, from main and not dev. 

The alive branch never needs to be synced, but can be synced manually. It can also be deleted at any time without issues.

Jon-b-m:
Ready for merging?

bjornoleh:
Yes

PR #558, opened 2024-02-29
Update HomeRootView.swift

nicolevanelstvandenhoek:
As you asked for 👍


Jon-b-m:
Almost, as you also changed the profile override image, which I don't think is necessary. 
I wanted to see what colour you were using (teal). If we're going to se this colour for highlighting the override button we should use same colour also for highlighting also the temp target button (which I also requested from you). If you can update your PR with these changes (revert of profile image and teal for highlighted temp target) I thing this PR can get merged!

Thank you!

PR #559, opened 2024-03-02
Step one towards adjustable Autotune Days

scrappy:
Add variable autotuneTuneDays to FreeAPSSettings.swift
    Move CarbsStorage.swift under Sources/APS/Storage
    Use autotuneTuneDays in:
      * storeCarbs()
      * storeGlucose()
      * storeEvents()
    Fix CarbsStorage.swift to match other *Storage.swift, by using settingsManager instead of settings

    All Changes are cosmetic ... default for autotuneTuneDays is still 1 day, this just cleans up the code
    a bit before moving to an adjustable autotuneTuneDays

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -1481,7 +1481,6 @@
 		3811DED425C9E1E300A708ED /* Resources */ = {
 			isa = PBXGroup;
 			children = (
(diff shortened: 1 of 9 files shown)


Jon-b-m:
Please don't PR half-done changes. Take your time and make a PR when complete, because it wouldn't make any sense merging this.

scrappy:
Will resubmit later, when complete ... everything seems to work *except* that I'm passing an array where it should be a json, so determineBasal fails ...

PR #595, opened 2024-03-20
contact trick

yurique:


yurique:
I'm going to add some details to the PR description (didn't have much time today). Also it still needs a pass or two of cleanup (for example, I suspect there are things that I put into the `ContactTrickState` that are not needed there, etc).

As of now, my biggest question would be this: am I doing things the right way in `BaseContactTrickManager`? (I copy-pasted the code from `WatchManager` initially, and then updated it).

Another thing is the list of things that can useful on a watch face. I added a bunch, but I've been using iAPS for just a couple of days, so I'm just guessing. Also need to double-check that the "stale" state is rendered correctly - with `--` instead of numbers, etc.

yurique:
Another thing - I called the thing "Contact Trick" everywhere, which might be not the best name/label :)

Jon-b-m:
* I think the naming is good, it's descriptive and intuitive. 

* The most important element to display is really the current glucose, and the reason for why the contact trick is even needed. 

After that I think green/orange/red loop. Perhaps even IOB as on option. 

The loop and the IOB and the COB and all of the other loop status can all be retrieved from just the suggestion. 

I therefore don't think we need all of the other stuff or imported modules here, as we don't need to enact anything (don't need to communicate with pump), just retrieve the last suggestion.

Jon-b-m:
I suggest to start small, then we/you eventually can add stuff later.

yurique:
Thank you for your thoughts on this, Jon!

I ended up setting up `BG`+`trend`+`IOB` on one contact and `green/orange/red loop`+`last loop time` on the second - and it does seem to work well:

IMG_3281

> I suggest to start small, then we/you eventually can add stuff later.

Sounds good! I'll be removing things/imports (later today, hopefully).

I also need to tweak the rendering a little bit, some things end up too small under certain circumstances - too hard to see inside a small circle.

yurique:
I think `last loop time` is important because the contact picture doesn't have an expiration time, unlike calendar events - so if something happens (iAPS crashes or something else) - the contact picture will stay there "forever" and will become misleading.

In fact, that's the biggest downside of this approach (and a risk), even though it hasn't happened a single time over the past two days of me running it. I think I should add some warning/description about this in the settings view.

yurique:
@Jon-b-m I pushed some tweaks-and-cleanups - removed most of the imports and subscriptions in the manager, dropped part of the "state" and tweaked the rendering a little bit.

I will need to try it "live" to see if it look good/is readable on the watch, but otherwise it all should be in a more or less decent state.

FreeAPS/Sources/Modules/Base/BaseProvider.swift
```diff
@@ -11,6 +11,7 @@ class BaseProvider: Provider, Injectable {
     @Injected() var deviceManager: DeviceDataManager!
     @Injected() var storage: FileStorage!
     @Injected() var bluetoothProvider: BluetoothStateManager!
+    @Injected() var contactTrickManager: ContactTrickManager!
```
yurique, line 14: not sure if I should have added this here - it's referenced from the `ContactTrick.Provider` when contacts are saved

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
```diff
@@ -0,0 +1,198 @@
...
+            self.state.eventualBG = self.eventualBGString()
+
+            self.renderContacts()
+        }
+    }
+
+    private func renderContacts() {
+        contacts.forEach { renderContact($0) }
+        workItem = DispatchWorkItem(block: {
+            print("in updateContact, no updates received for more than 5 minutes")
```
yurique, line 75: haven't looked yet how to log things like this correctly

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
```diff
@@ -0,0 +1,198 @@
...
+        storage.retrieve(OpenAPS.Enact.suggested, as: Suggestion.self)
+    }
+}
+
+extension BaseContactTrickManager:
+    GlucoseObserver,
+    SuggestionObserver,
+    SettingsObserver
+{
+    func glucoseDidUpdate(_: [BloodGlucose]) {
```
yurique, line 187: not sure if it's necessary to subscribe to glucose updates, or if subscribing to suggestions is enough

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
```diff
@@ -0,0 +1,198 @@
...
+        let units = settingsManager.settings.units
+        return eventualFormatter.string(
+            from: (units == .mmolL ? eventualBG.asMmolL : Decimal(eventualBG)) as NSNumber
+        )!
+    }
+
+    private var glucoseFormatter: NumberFormatter {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        formatter.locale = Locale(identifier: "en_US")
```
yurique, line 152: I added the locale explicitly here in order to make it print decimal separators as points (`.`) not commas (`,`), because It looked weird-ish on the complication, but it doesn't seem right to override the locale like this 🤔

FreeAPS/Sources/Modules/Settings/View/SettingsRootView.swift
```diff
@@ -61,6 +61,9 @@ extension Settings {
                     Text("Dynamic ISF").navigationLink(to: .dynamicISF, from: self)
                 } header: { Text("Extra Features") }
 
+                Section {
+                    Text("Contact trick").navigationLink(to: .contactTrick, from: self)
```
yurique, line 65: right now the Contact Trick section is located in the main settings view, right above the `Debug options` - not sure what's the right place for it

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
```diff
@@ -0,0 +1,196 @@
...
+    private func renderContacts() {
+        contacts.forEach { renderContact($0) }
+        workItem = DispatchWorkItem(block: {
+            print("in updateContact, no updates received for more than 5 minutes")
+            self.renderContacts()
+        })
+        DispatchQueue.main.asyncAfter(deadline: .now() + 5 * 60 + 15, execute: workItem!)
+    }
+
+    private func renderContact(_ entry: ContactTrickEntry) {
```
yurique, line 79: this larger functions a straight copy-pastes from the `WatchManager`, it probably deserved a refactoring but I don't feel qualified for that :)

FreeAPS/Sources/Services/ContactTrick/ContactPicture.swift
```diff
@@ -0,0 +1,845 @@
...
+        state: ContactTrickState,
+        rect: CGRect,
+        fitHeigh: Bool,
+        fontName: String?,
+        fontSize: Int,
+        fontWeight: UIFont.Weight,
+        fontTracking: FontTracking,
+        color: Color
+    ) {
+//        guard let context = UIGraphicsGetCurrentContext() else {
```
yurique, line 264: commented out - just some stuff to help debugging, to be deleted

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
```diff
@@ -0,0 +1,196 @@
...
+            self.renderContacts()
+            completion(.success(()))
+        }
+    }
+
+    private func configureState() {
+        processQueue.async {
+            let readings = self.coreDataStorage.fetchGlucose(interval: DateFilter().twoHours)
+            let glucoseValues = self.glucoseText(readings)
+            self.state.glucose = glucoseValues.glucose
```
yurique, line 53: `self.state` doesn't have to be an instance field in this class - I could create a new one every time and pass it along into `renderContacts()` - but it was done like this in `WatchManager` so I kept it the same

yurique:
I was observing something just now - the ring was green, last update was 3 min ago, but on the contact picture it was yellow (and 8 minutes ago) - I did "force" the refresh of the contact, so it seems like my code was getting a wrong `time of the latest cycle` somehow?

```swift
self.state.lastLoopDate = self.apsManager.lastLoopDate
```

Jon-b-m:
> I was observing something just now - the ring was green (in the app), last update was 3 min ago, but on the contact picture it was yellow (and 8 minutes ago) - I did "force" the refresh of the contact, so it seems like my code was getting a wrong `time of the latest cycle` somehow?
> 
> ```swift
> self.state.lastLoopDate = self.apsManager.lastLoopDate
> ```
> 
> (it got fixed with the next cycle, so I didn't have a chance to investigate more)

You don’t need APSmanager for this. Everything can be retrieved from suggestion. suggestion.timestamp

Jon-b-m:
I will take a closer look at this PR next week, I hope. As soon as I’m done with some other iAPS work. Thank you!

yurique:
> You don’t need APSmanager for this. Everything can be retrieved from suggestion. suggestion.timestamp

Aha! Very good! :)

yurique:
Another thing I'm observing today - the contact pictures on the watch face are displaying the pictures from 5 hours ago, even though the contact manager was updating the contacts without problems. Tapping on the complication opens the contact - the picture is up-to-date there, as well as when looking at the contact on the phone. Not sure what has happened. Locking/unlocking the watch fixed it.

yurique:
I had to restart the watch to fix it. The contact pictures on the watch were not updating, or updating but with a very long delay.

FreeAPS/Sources/Services/ContactTrick/ContactTrickManager.swift
yurique: I guess BG can also be retrieved from the suggestion, instead of asking `self.coreDataStorage.fetchGlucose(...)`

yurique: but not `trend` or `delta` 🤔

yurique:
A quick update: I haven't seen any issues with contacts being out of date on the watch face since the last time I mentioned it here - works like clock.
I suspect it might have been due to the bug I had in the code - there is a task that is scheduled to run 5+ minutes after the last update (if no other updates received), and somehow I managed to delete the line that would cancel the previously scheduled task when needed. So after a while, the contacts update would end up running a LOT.

Jon-b-m:
Now merged into dev via another branch, as I had to make some small edits after testing. Thank you @yurique

PR #622, opened 2024-04-28
Clean Oref0 and new charts

Jon-b-m:
* Clean Oref0
- Use the OpenAPS Oref0 repo modules unmodified. The algorithm in JS is back to the Oref0 modules, as originally implemented in FreeAPS.
- The source code for dynISF is added as a layer and run before the OpenAPS algorithm, similar to the middleware, visible and uncompressed in the iAPS repo. The TDD code needed for dynISF now written in Swift, made to work for all pumps using the pump history storage, as before.
- One exception where the JS code currently differs from OpenAPS is in one file, a file just with the default list of preferences etc, the index.js file. This is taken compressed from the original FreeAPS code (https://github.com/ivalkou/freeaps/blob/master/FreeAPS/Resources/javascript/bundle/profile.js), as this code has never been presented unminimized and I therefore don’t know exactly what format is used for this file, but you can see all these variables contained here uncompressed in the preferences.json and in other places in the Swift code, where this list of preferences and profiles is generated. 

* Bolus Calculators
- Both the “original” and the alternative are now written entirely in Swift. Several bugs found and fixed in the newer alternative calculator. The calculators now about 0.2 s faster, when not using the slower LoopKit function for rounding.
- Added safety guardrails to both alternatives, using the OpenAPS prediction minPredBG (can be disabled, but on by default). The error alerts have been removed and replaced with illustrations.
- Settings for bolus calculators have been  revamped. A bit simpler now, I hope.

* New Charts
- New charts and illustrations for the bolus calculators. You can now see the insulin needed illustrated in the bolus prediction chart as a AreaMark in the prediction chart in bolus view.
- Target glucose added as dotted green line.
- For the alternative calculator I added new graphics to better illustrate the data used, a new chart with the insulin needed for each variable and a Grid data table with the specific formulas and summation of the variables.
- Main Chart. Alignment of the default bolus and carbs dots and strings adjusted for easier reading of all entries. To reduce eventual bolus string clutter a setting has been added in UX/U settings, where you can opt in to hide the insulin amount string when under a selected value, default is 0.3, a bit similar to xDrip. As a test I’ve added an option in UI/UX-settings to replace the bolus dot with an InsulinBar instead (off by default). The height of the bars is dynamic, based on your pump history, meaning a user with history of boluses of 0.2-2 U will generate the same size of the insulin bars as a user with history of boluses of 2-20. Max Height is added as a new Config in code and will impact all bars.

* Miscellaneous
- Middleware and the adjusted profiles can now be read outside of the middleware layer and outside of oref0.
- It’s now clear when autosense has been used and when the dynamic ratio has been used instead.
- TDD will always be displayed in the history data table. When dynamic ratio used also in the iAPS loop pill (pop-up). You can see hours of data available in the data table.
- The dynISF and the other OpenApS variables are now separated in two groups in the pop-up.
- Any active override is now added to the iAPS pill pop-up and to the Nightscout OpenAPS pill. Middleware can be read in pop-up and NS pill as before.
- The profile button was slightly disproportionate (too large) before.
- Changed colour of glucose in main chart to darkGreen, for better readability and to be coherent with the TIR and Loop chart green colour.

Tested for more than 6 weeks in vivo.


MikePlante1:
“Tested for more than 6 weeks in vivo” is a very misleading statement when you posted “I found a bug” in the v4 Discord channel a week ago.

Jon-b-m:
Your comment is borderline hostile,  as you don’t even eant to know what bug I actually found and fixed? 

The bugs were in the older alternate bolus calculators code which dates back to November 2013 (tested since 5 months ago both by main and dev users). 

And yes I’ve been testing my new commits for more than 6 weeks in vivo. During my testing I also found other older bugs which I fixed, not related to my new commits but still needing fixing, since I’m the manager of this repo and which will be important for all users.

PR #663, opened 2024-05-08
Contact Image - font weight, font tracking, preview background fill

yurique:


yurique:
merged as part of the other PR, closing

PR #672, opened 2024-05-10
fix importing settings from Nightscout

MikePlante1:
Before this PR, two settings were not imported from Nightscout correctly for me.
My 95 mg/dL target glucose got imported as 72 mg/dL (the lowest allowed)
My 58mg/dL ISF got imported as 9 mg/dL

After applying this PR and building to a fresh now simulator, my settings were imported correctly.

### DIFF
--- a/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
+++ b/FreeAPS/Sources/Modules/NightscoutConfig/NightscoutConfigStateModel.swift
@@ -212,7 +212,7 @@ extension NightscoutConfig {
 
                         let sensitivities = fetchedProfile.sens.map { sensitivity -> InsulinSensitivityEntry in
                             InsulinSensitivityEntry(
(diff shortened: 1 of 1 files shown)


Jon-b-m:
Ok, did you test with both configured glucose units in iAPS?

MikePlante1:
I only tested with mg/dL in iAPS (latest dev, and then latest dev + this PR)

I did test mmol/L in a different fork, and it worked fine after applying the PR as well, but I used someone else's NS account for that so I couldn't use it to test iAPS.

Jon-b-m:
So this needs some more testing by more users, with both units.

MikePlante1:
Should be a pretty quick test in a simulator if you find some willing testers.

For mmol/L, though, the logic is unchanged. The only changes in this PR are replacing the 3 ternary operators with the default for mmol/L users.

Jon-b-m:
I've tested now. As mentioned this only affect mg/dl users. 
Thanks for PR. Will merge

PR #689, opened 2024-05-18
autofat changes to UI. Still needs to be wired in look at FPU fat ent…

rsilvers129:
Added the UI for autofat including saving states and needed settings. Still needed: Wire in to reduce the bolus based on the FPU fat entry.


Jon-b-m:
Thanks for PR.
1. You need to remove your signing (5RLXQ74459) from this PR. If you use the ConfigOverride.xcconfig instead or just don't stage your signing before committing.
2. You inadvertently removed the default "displayPredictions" : true variable. 
3.  The  HStack:

               HStack {Text(state.fattyAutomaticToggle ? "Automatic Setting" : "Manual Setting")
                                 Spacer()
                                 Toggle("", isOn: $state.fattyAutomaticToggle)
              }

 you could convert to a Toggle instead:

               Toggle(isOn:  $state.fattyAutomaticToggle) {
                                  Text(state.fattyAutomaticToggle ? "Automatic Setting" : "Manual Setting")
                }

but probably best to change the label so that users understand that the toggle adjusts two different settings. Perhaps "Automatic / Manual" would be more appropriate?

PR #722, opened 2024-05-30
Fix birthDate default / typo

scrappy:
* Fix Typo: birtDate -> birthDate
* Set default birthDate to distantPast ( current default is "today", making everyone an infant )

I tried using distantFuture, but on my simulator, that ended up being May 1, 2024 ... which is the past, not future, which made no sense as the one thing I found online seemed to indicate it should be something like year 4001 ... but the docs themselves show that distant{Past,Future} should both be 'off by centuries', so how it came with this year, no clue.

distantPast is slightly better ... it comes to Jan 1, 1920 ... still not "centuries in the past", but hopefully that is a fixed number, and not floating, so I can filter for that date, with a hopefully pretty good certainty that there aren't a lot of 104 year old users or older that is using iAPS that I am eliminating ... can't say for sure, but I think odds are pretty good ...

### DIFF
--- a/FreeAPS/Resources/json/defaults/freeaps/freeaps_settings.json
+++ b/FreeAPS/Resources/json/defaults/freeaps/freeaps_settings.json
@@ -62,5 +62,5 @@
   "minimumSMB": 0.3,
   "useInsulinBars": false,
   "uploadVersion": true,
-  "birtDate": new Date()
+  "birthDate": Date.distantPast
 }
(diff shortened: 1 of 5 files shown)


scrappy:
Okay, just figured out why Jan 1, 1920 ... I'm setting the value to some day in "way past", but DatePicker itself only goes back *to* 1920, based on dateRange ... and that is why the distantFuture came out weird, since the range is "1920 to this month" ...

the code makes sense now ... unless I've missed something, please commit this one as it will fix the 'infant' situation ...

Merkaffe:
Yes, the range is from 1920 to now (this month).

PR #734, opened 2024-06-09
[DanaKit] Progress update 09-06-2024

bastiaanv:
Critical bugfix:
- fix for double insulin entries


bastiaanv (commented):
hot fix

Jon-b-m:
Is the force unwrapping in line 358 absolutely necessary?

bastiaanv:
I could rework it to the following:

```swift
            if timeResult.success {
                let data = self.state.usingUtc ? (timeResult.data as? PacketGeneralGetPumpTimeUtcWithTimezone): (timeResult.data as? PacketGeneralGetPumpTime)
                guard let date = data?.time else {
                    self.log.error("Invalid package received")
                    return nil
                }

                return date
            } else {
```

PR #773, opened 2024-07-16
Update Localizable.strings - Sharing section

Mirko-tri:
Found non translated strings in my App (Main branch).
Added those from the "sharing" section.
First time doing this ¯\_(ツ)_/¯


Merkaffe:
Looking good, thanks!

PR #805, opened 2024-08-08
Use original oref0 profile script

thomasvargiu:
Use the original `lib/profile/index.js` from [`oref0`](https://github.com/openaps/oref0).

I can't be 100% sure whether there are other changes or not, but this PR handles the errors like the actual bundled one, using the original script.

Changes:
- changed `scripts/webpack.config.js` to use original files
- changed `scripts/webpack.config.js` to allow another directory for oref0 lib using an env variable
- changed `scripts/webpack.config.js` with optimizations flags for sideEffects and comments
- added `package.json` to lock webpack dependency

When someone wants to build the oref0 lib, they can:
```
$ git clone https://github.com/openaps/oref0.git
$ npm install
$ npm run build
```


Jon-b-m:
I've tested the for some time now. Everything seems fine. the comment about current dev/main index.js is not accurate (but not very important), as all variables is included here (you can check for yourself with for instance a search in Xcode), power your solution is prettier and highly appreciated as we now can use every files from the public official ore0 repo!

thomasvargiu:
I was thinking to refactor the other files prepare files. It should be better to don't use the workaround to inject custom iAPS variables in the profile. They are not for oref0 and we should get them from the preferences where we need them.

Jon-b-m:
> I was thinking to refactor the other files prepare files. It should be better to don't use the workaround to inject custom iAPS variables in the profile. They are not for oref0 and we should get them from the preferences where we need them.

I'm open to this. There are several different ways of refactoring we can make.

Jon-b-m:
This PR is marked as “draft”, do you have additional updates pending?

thomasvargiu:
It's a workaround, but it works...
I would like to change how custom iAPS settings are handled (currently merged in profile.json, even they are not used my oref0), but it will require to change how the middleware works causing a breaking change.
What do you think? Maybe we can merge this and maybe later change how custom preferences are handled?

Jon-b-m:
> It's a workaround, but it works... I would like to change how custom iAPS settings are handled (currently merged in profile.json, even they are not used my oref0), but it will require to change how the middleware works causing a breaking change. What do you think? Maybe we can merge this and maybe later change how custom preferences are handled?

Braking changes are not ideal, and should only be introduced when absolutely necessary. What changes to the middleware were you planning?

thomasvargiu:
> Braking changes are not ideal, and should only be introduced when absolutely necessary. What changes to the middleware were you planning? 

Removing the custom iAPS settings from the profile and move them to preferences and provide them to the middleware.
Another improvement to the middleware should be to use just one argument as an object.

But we can think about these things in the future. I'm removing the draft flag from this PR.

Merkaffe:
> > Braking changes are not ideal, and should only be introduced when absolutely necessary. What changes to the middleware were you planning?
> 
> Removing the custom iAPS settings from the profile and move them to preferences and provide them to the middleware. Another improvement to the middleware should be to use just one argument as an object.
> 
> But we can think about these things in the future. I'm removing the draft flag from this PR.

Are you referring to the  dynamic ISF settings? These are already in preferences.json. But I agree it would be more structured  to move these to freeaps_settings.json instead as these are not used in oref0.  We could do this without breaking the middleware for most users, unless for those users trying to read these settings from preferences object in middleware (there is no good reason to as these are part of profile.json).

And yes I think this PR can be merged for now and the other refactoring happening later in a separate PR. I've been testing this PR thoroughly.

PR #876, opened 2024-09-19
Allow for non-default Bundle ID

scrappy:
Tested with both APP_IDENTIFIER set and not set

To test, go into Settings -> Secrets and Variables -> Actions -> Variables tab -> New Repository Variable

Name: APP_IDENTIFIER
Value: new bundle id

The code creates an appropriate ConfigOverride.xcconfig file in the work directory for the build

This code can easily be extended for anything else that would be useful to be in that file


Jon-b-m:
Tested successfully.
For other users beware that the new variable is not a new secret, but a repository variable.

PR #926, opened 2024-11-05
Add (ONE+) & (ONE) to the display name

i3bdulr7man:
This PR simply updates the name of Dexcom G7 and Dexcom G6 to "Dexcom G7 & ONE+" and "Dexcom G6 & ONE" in the CGM selection menu, making it clearer for Dexcom users to identify the correct option.


Jon-b-m:
Did you test this PR?

i3bdulr7man:
> Did you test this PR?

The modification made to the code is simple, and I don’t think it requires testing.

Jon-b-m:
> > Did you test this PR?
> 
> The modification made to the code is simple, and I don’t think it requires testing.

With the empirical data you would also see that this PR doesn't work.

PR #1002, opened 2024-12-14
autotune view - show current basal rates

yurique:
<details>
  <summary>screenshot:</summary>
  
image

</details>

### DIFF
--- a/FreeAPS/Sources/Modules/AutotuneConfig/AutotuneConfigProvider.swift
+++ b/FreeAPS/Sources/Modules/AutotuneConfig/AutotuneConfigProvider.swift
@@ -15,5 +15,11 @@ extension AutotuneConfig {
         func deleteAutotune() {
             storage.remove(OpenAPS.Settings.autotune)
         }
+
+        var profile: [BasalProfileEntry] {
+            storage.retrieve(OpenAPS.Settings.basalProfile, as: [BasalProfileEntry].self)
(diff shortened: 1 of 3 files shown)


Jon-b-m:
Thanks. I think it would look slightly cleaner without seconds  in first column and with the second column (…->) aligned as leading.

yurique:
I'll give it a shot!

yurique:
I tried changing the `rateFormatter` to always include 2 fraction digits - it did help align the numbers a bit, but it's still not perfect (since it's not a fixed-width font).

Otherwise I'm not sure how to make the whole thing to behave like a table (so that I could align the "columns" perfectly), without making the widths fixed - which would probably be a bad idea.

Here's how it looks like:

<details>
  <summary>screenshot</summary>

image

</summary>

Jon-b-m:
Looks good 👍. Yes iOS 16 is required. And iAPS has some SwiftUI code which is only used when using iOS 17+.

Jon-b-m:
Do we perhaps need a total also? To compare the total autotuned with normal rates?

yurique:
Makes sense!
 I'll add it a bit later.

yurique:
<details>
  <summary>show current total</summary>

image

</details>

<details>
  <summary>seems to match what's displayed in the profile editor</summary>

image

</details>

Jon-b-m:
Accepted.

PR #1007, opened 2024-12-20
live activity - chart

yurique:
Adding charts to the live activity view (lock screen).

Adding two toggles to the Notification settings section:
* Display Chart
* Show Predictions

The chart displays:
* latest 2 hours of readings
* up to 2 hours of predictions (if enabled)
* the high and low threshold lines

Things that could be easily tweaked:
* the size of the points on the chart 
* the period to display (currently 2 hours)

Ideas I have not explored:
* show SMB markers?

Screenshot with predictions:
image

Screenshot without predictions:
image


yurique:
Pushed some changes after the discussion in Discord.
This is not final - I added a couple of toggles to the Notifications settings so we can try out different layouts.
<details>
  <summary>screenshots:</summary>

####  Settings:
  image

#### Small status circle, small delta+eventual:
image



#### Original (large) status circle, small delta+eventual
image


#### Small status circle, eventual at the bottom, no delta
image


#### Large status circle, eventual at the bottom, no delta
image


</details>

Jon-b-m:
7 settings for a small live activity widget is a bit excessive. Only setting needed, if any, is display chart.

yurique:
> 7 settings for a small live activity widget is a bit excessive

I totally agree.
All these settings are here so that we can test it all on a real device, to see what works/looks best, most of them will be removed before merge. 

As for which of them to leave in the final version - I'm not sure. It looks like (from the discussion in Discord) different people have different preferences about how they want it to look like:
* you prefer to have the timestamp at the top - it makes sense because it's the standard location for most (all) notifications/badges; but it takes away vertical space. (I tried moving it into the first row, but this makes the first row rather crowded/cramped)
* you mentioned that the option with eventual glucose at the bottom looked the cleanest - it does look not bad, but it also takes a lot of space from the chart, so I personally would prefer it to be to the right of the chart
* if eventual is to the right of the chart - I'd prefer the loop time to be there as well, at the bottom, to to free some space at the top
* I prefer to have the threshold lines visible, but I'm not sure everyone else wants that as well 
...etc

I'm currently replacing a couple of toggles with a single picker (I'll call it "chart layout" or something like that). But again - this is not meant to be the final version - only to see what works better on the phone.

yurique:
Some thoughts on the settings in the final version:
* we can make the chart always on - so no toggle for the chart
* `Arrow for Eventual Glucose` - it's not really related to charts (it changes the text "Eventual Glucose" in the bottom row into an `⇢` - it's probably not really needed if the chart is always on
* `Small Status Circle` - this one is probably not needed either, we just need to decide where we want the status circe: keep as it is now - large circle on the left, or move it next to the timestamp
* `Show Predictions`, `Threshold Lines` - to me it seems like these are something people might want to customize (depending on their screen size, etc)
* `Dynamic Range` - I think it's okay to remove it and make it always be "dynamic" ("dynamic" means the range of the chart depends on the actual reading, vs fixed hardcoded values like `3.0mmol/l .. 22.0mmol/l`)

The new `Layout` picker - I think it's going to be the most important/tricky one. If we can find a layout that everyone is happy with, we can drop it. But can we? :) 

---
Here's the current settings and layouts:
<details>
  <summary>screenshots</summary>


#### Settings:


#### 1. Eventual BG on the bottom with small circle



#### 2. Eventual BG on the top with large circle



#### 3. Eventual BG on the right with large circle



#### 4. Eventual BG on the right with small circle



#### 5. Eventual BG on the right with large circle and timestamp in the bottom 



#### 6. No Eventual BG with large circle
image


</details>

yurique:
Pushed an update removing most of the settings and different layouts. Updated the description to reflect the current state.

Jon-b-m:
Adding a glucose chart is a nice idea, especially when making it optional. However compared to the normal View (current dev) it is a bit harder to read with current UI suggestion. I think we perhaps need to rearrange the entire View when adding the chart to make it more readable/cleaner. Adding a chart background (or backgrounds) would probably help making the added data look less scrambled together?

PR #1048, opened 2025-01-17
DecimalTextField - "listen" and update to value changes when not editing

yurique:


FreeAPS/Sources/Views/DecimalTextField.swift
```diff
@@ -90,6 +91,12 @@ public struct DecimalTextField: UIViewRepresentable {
     }
 
     public func updateUIView(_ textField: UITextField, context: Context) {
+        if !context.coordinator.isEditing {
+            let newText = valueAsText()
+            if textField.text != newText {
+                textField.text = newText
+            }
+        }
```
yurique, line 99: this is the actual fix, everything else is just a small cleanup/refactoring

PR #1049, opened 2025-01-17
DecimalTextField - live editing mode

yurique:


FreeAPS/Sources/Views/DecimalTextField.swift
```diff
@@ -163,6 +168,21 @@ public struct DecimalTextField: UIViewRepresentable {
...
+            let proposedText = textField.text ?? ""
+
+            if let number = parent.formatter.number(from: proposedText) {
+                let decimalNumber = number.decimalValue
+                parent.value = decimalNumber
+            } else {
+                // invalid input, set value to 0
+                parent.value = 0
+            }
+        }
```
yurique, line 185: the main change here

FreeAPS/Sources/Views/DecimalTextField.swift
```diff
@@ -85,7 +88,9 @@ public struct DecimalTextField: UIViewRepresentable {
             action: #selector(Coordinator.cancelEdit)
         )
 
-        toolbar.items = [clearButton, flexibleSpace, cancelButton, doneButton]
+        toolbar
+            .items = liveEditing ? [clearButton, flexibleSpace, doneButton] :
+            [clearButton, flexibleSpace, cancelButton, doneButton]
```
yurique, line 93: don't show Cancel when liveEditing=true since the value is already updated and we can't cancel

FreeAPS/Sources/Modules/DataTable/View/DataTableRootView.swift
```diff
@@ -315,7 +315,7 @@ extension DataTable {
                                     value: $state.externalInsulinAmount,
                                     formatter: insulinFormatter,
                                     autofocus: true,
-                                    cleanInput: true
+                                    liveEditing: true
```
yurique, line 318: here and in other places `cleanInput: true` is the default, so I removed it, and added `liveEditing: true`

in the diff it looks like just a change of the parameter name

PR #1067, opened 2025-01-23
live editing in all profile override fields

yurique:
The user can tap the "Start" button without leaving the text field they're editing which can (potentially) be a problem:
*  the user will think they changed the value, but it will not be saved


Jon-b-m:
Thanks. Is it tested with real device?

yurique:
> Thanks. Is it tested with real device?

Yes, I've tested it on my phone.

PR #1121, opened 2025-02-04
Draft: testing webview to execute Javascript

thomasvargiu:
The idea is to use a WebView to execute Javascript and take it in memory. MKWebView uses JIT, it's much faster for our needs.

I'm not a Swift Developer and I can't continue on it, so this MR is just a test but could be a good starting point if someone want to try and work on it.

In my case, testing it, the loop is usually 2x/3x faster.

AISF is not migrated in the WebView yet. Custom scripts should be isolated because the idea is that everything is loaded and we can't accept variable and functions collisions.

@Jon-b-m does it worth some experiments?


thomasvargiu:
It should work, but I didn't test it honestly. Probably the better solution when including middleware and other "custom" javascript, would be to include it with something like this in order to isolate it:
```swift
let script = """
var middleware = (function () {
    \(script)
    
    return \(generate_function);
})()
"""
```

Jon-b-m:
OK. I know it's still labeled as a draft, but I did a test in simulator. It works when dynamic and AISF is disabled.

thomasvargiu:
I've updated the code. You can test it. I'm testing it with dynamic and AISF.

Now every script is isolated.

In the code, I needed to remove the iasf script loaded from storage. Can you tell me if it's really necessary and why can be overrided?

Jon-b-m:
Thank you. It looks better now 👍. It’s not strictly necessary, it was just a matter of refactoring, however it’s better this way (your suggestion).

Jon-b-m:
How do you mean overridden?

Jon-b-m:
Testing. now again. it crashes on the first loop. 
image
I can do more tests tomorrow.

thomasvargiu:
@Jon-b-m ok, seems WKWebView needs to be called from main thread. I've updated it in order to execute it on the main thread (even if I wasn't receiving that error).
In this new commit I've also added a way to fetch files from storage in a parallel way. I think it's working but since a DispatchQueue extension is used and, I repeat, I'm not a Swift developer, I'm not 100% sure that file reads are 100% thread safe.

This branch must be considered an experimental test and if we want to procede, someone with experience should make the changes :)

Jon-b-m:
I will look at this more closely later. So far I’ve only run a couple of builds in sim as I was curious. I agree this is highly experimental

thomasvargiu:
Well, an example of debugging on my iPhone 14 Pro:

`dev` branch:
```
Time for Loading files 0.33324599266052246 seconds
Time for Meal module 0.4633030891418457 seconds, total: 0.7965760231018066
Time for IOB module 0.2923499345779419 seconds, total: 1.0959349870681763
Time for Middleware module 0.07364296913146973 seconds, total: 1.1696009635925293
Time for AutoISF module 0.13382399082183838 seconds, total: 1.303449034690857
Time for Determine Basal module 0.13776397705078125 seconds, total: 1.4412330389022827
```

`js-webview` branch:
```
Time for Loading files 0.2913200855255127 seconds
Time for Meal module 0.05555105209350586 seconds, total: 0.3469020128250122
Time for IOB module 0.04194200038909912 seconds, total: 0.39671194553375244
Time for Middleware module 0.010918974876403809 seconds, total: 0.4076499938964844
Time for AutoISF module 0.0318220853805542 seconds, total: 0.43956899642944336
Time for Determine Basal module 0.020382046699523926 seconds, total: 0.46031200885772705
```

Jon-b-m:
That is a good start! Interestingly there's some extra second while in bolus View I need to look at also later. Could be some SwiftUI/animation glitch here...

Jon-b-m:
No, not SwiftUI. It's the APSManager I think

Jon-b-m:
Middleware works. The FileStorage is the bottleneck now. If using CoreData we would save 0.3 seconds more.

Jon-b-m:
We can create the JSONs without using Disk.

thomasvargiu:
I was trying to refactor the storage right now, even using a dispatch queue for each file in order to have the best performance

Jon-b-m:
After running it for around 12 hours no weird issues. The cpu and memory looks Ok compared to before in debug views.

Jon-b-m:
Average loop duration compared to last month and year is 1.8 seconds less.

thomasvargiu:
Me neither for 24 hours, until now, but I've made more changes not yet in this branch:


image

thomasvargiu:
Probably not related to this branch. It started working again after 1 hour.

Jon-b-m:
We don’t know without some more context. It could be a stale or flat glucose or other normal reasons. It will probably be apparent from looking at the logs and chart.

Jon-b-m:
> It was probably my bug, but right now I've made some other improvements:
> 
> Before
> 
> ```
> Time for Determine Basal module 0.13776397705078125 seconds, total: 1.4412330389022827
> ```
> 
>  ```
> Time for Determine Basal module 0.028442978858947754 seconds, total: 0.21813499927520752
> ```
> 
> **From ~1.2 seconds to 0.2 seconds!**
> 
> Mostly because we were fetching from CoreData `fetchInsulinDistribution` without a limit but using just the first value (I have ~120.000 records).

Very good 👍 I’ve overlooked this. Btw why do you initialize the CoreDataStack several times? Shouldn’t be needed?

Jon-b-m:
I advice not to use ChatGPT as it can be a parrot repeating bad advice/bad coding practices.

thomasvargiu:
Of course, it's not for coding, it's just for documentation on iOS and Swift programming. I'm aware that using it for coding it's a bad idea

thomasvargiu:
Sometimes, with this branch, I have error on decoding JSON from javascript functions. No other errors from logs. After restarting the app, it works. I guess using an "always-up" WKWebView can cause some issues that need to be handled.

Jon-b-m:
> Sometimes, with this branch, I have error on decoding JSON from javascript functions. No other errors from logs. After restarting the app, it works. I guess using an "always-up" WKWebView can cause some issues that need to be handled.

Which errors explicitly?

Jon-b-m:
Updated branch to alpha, as I think it's better to test something as substantial as this in alpha first.

thomasvargiu:
> > Sometimes, with this branch, I have error on decoding JSON from javascript functions. No other errors from logs. After restarting the app, it works. I guess using an "always-up" WKWebView can cause some issues that need to be handled.
> 
> 
> 
> Which errors explicitly?

Honestly I don't know. I looked at the logs, and none of the JavaScript functions was working (for a lot of time), every call returned an error for an invalid json.
Restarting the app resolved the problem.

I suggest you to create a new branch if you want to test it, in this branch there are too many unrelated changes and should be closed IMHO.

Jon-b-m:
OK. Your PR consists of two parts: async file processing and the wkwebview. I don't see any other changes?

Jon-b-m:
But yes probably best to test each separately.

Jon-b-m:
@thomasvargiu Do you have any updates / refactoring of this PR planned for or in the pipeline?

thomasvargiu:
@Jon-b-m it was just a test branch. Async file processing probably isn't necessary. I suggest you to open a new branch/MR to implement just the wkwebview. I'm sorry but I don't have time right now.

I'm a little worried about the problem I had when I was using this branch, but we can have more people who can test it if we'll merge it on an alpha branch.

Jon-b-m:
OK. I’ve been able to save 0.3 s more by doing some more asynchronous file processing in my current local repo. 
The async does make a difference, however I’ve never tested just running the wkwebview. It would be useful seeing  exactly what makes the time saving. Is it the async file processing, the wkwebview or both and in what proportion?

Jon-b-m:
I've also resolved all Swift6 warnings your PR was generating in my local repo. I can probably merge this local repo later to alpha...

Jon-b-m:
I'm closing this PR anad merging the local changes into alpha now.

PR #1124, opened 2025-02-05
profiles summary: tweaks

yurique:
Changing the text for SMB delivery range:
* if any of the settings are changed (min, max, range) - the whole description is displayed
* instead of three separate texts - display one formatted text like `0.5-0.7 range 5.8 mmol/L`
* if min=max, display just one number

---

Duration is formatted: * `1h 3m`.

---

Use the number formatter for decimal values, to prevent it from displaying ugly floating point numbers like `1.7898989898989898989`

---

Refactoring to use `if something { Text(something) }` instead of `something ? Text(something) : nil` - adding `nil`s seemed to affect the layout (like adding extra pixels of padding)

Before/after compared to `dev`:

---

image

image

---

image

image

### DIFF
--- a/FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
+++ b/FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
@@ -560,10 +560,7 @@ extension OverrideProfilesConfig {
                             if !isEditingPreset {
                                 isSheetPresented = true
                             } else if let editThis = presetToEdit {
-                                if !editThis.hasChanges {
-                                    moc.delete(editThis)
-                                }
-                                state.savePreset()
+                                save(editThis)
                                 isEditingPreset.toggle()
                             }
                         }
@@ -637,9 +634,6 @@ extension OverrideProfilesConfig {
             let scheduledSMBstring = (preset.smbIsOff && preset.smbIsAlwaysOff) ? "Scheduled SMBs" : ""
             let smbString = (preset.smbIsOff && scheduledSMBstring == "") ? "SMBs are off" : ""
             let targetString = targetRaw > 10 ? "\(glucoseFormatter.string(from: target as NSNumber)!)" : ""
-            let maxMinutesSMB = (preset.smbMinutes as Decimal?) != nil ? (preset.smbMinutes ?? 0) as Decimal : 0
-            let maxMinutesUAM = (preset.uamMinutes as Decimal?) != nil ? (preset.uamMinutes ?? 0) as Decimal : 0
-            let maxIOB = preset.overrideMaxIOB ? (preset.maxIOB ?? 999) as Decimal : 999
             let isfString = preset.isf ? "ISF" : ""
             let crString = preset.cr ? "CR" : ""
             let basalString = preset.basal ? "Basal" : ""
@@ -660,8 +654,8 @@ extension OverrideProfilesConfig {
                             .foregroundStyle(.secondary) : nil
                         smbString != "" ? Text(smbString).foregroundStyle(.white.opacity(0.8)).boolTag(false) : nil
                         scheduledSMBstring != "" ? Text(scheduledSMBstring).foregroundStyle(.secondary) : nil
-                        if let settings = autoisfSettings, settings.autoisf != state.currentSettings.autoisf {
-                            Text("Auto ISF \(settings.autoisf)").foregroundStyle(.secondary)
+                        if let aisf = autoisfSettings, preset.overrideAutoISF, aisf.autoisf != state.currentSettings.autoisf {
+                            Text("Auto ISF: \(aisf.autoisf)").boolTag(aisf.autoisf)
                         }
                         Spacer()
                     }
@@ -673,15 +667,17 @@ extension OverrideProfilesConfig {
                             percent != 1 && !(preset.isf && preset.cr && preset.basal) ? Text("Adjust " + isfAndCRstring) :
                                 nil
                             if !preset.smbIsOff {
-                                Text(maxMinutesSMB == 0 ? "" : maxMinutesSMB.formatted() + " SMB")
-                                Text(maxMinutesUAM == 0 ? "" : maxMinutesUAM.formatted() + " UAM")
+                                decimal(decimal: preset.smbMinutes ?? 0, setting: state.defaultSmbMinutes, label: "SMB ")
+                                decimal(decimal: preset.uamMinutes ?? 0, setting: state.defaultUamMinutes, label: "UAM ")
                             }
-                            maxIOB != 999 ? Text("Max IOB: " + maxIOB.formatted()) : nil
+                            if preset.overrideMaxIOB {
+                                decimal(decimal: preset.maxIOB, setting: state.defaultmaxIOB, label: "Max IOB: ")
+                            }
                         }.foregroundStyle(.secondary).font(.caption)
                     }
 
                     // All of the Auto ISF Settings (Bool and Decimal optionals)
-                    if let aisf = autoisfSettings, aisf.autoisf {
+                    if preset.overrideAutoISF, let aisf = autoisfSettings, aisf.autoisf {
                         let standard = state.currentSettings
 
                         HStack {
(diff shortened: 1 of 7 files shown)


Jon-b-m:
After you're done please make a screenshot before after (comparing with latest dev at that time) to illustrate what you want to accomplish. As mentioned elsewhere there already is a PR in progress, why I suggest looking at this also, to avoid extra work.

Jon-b-m:
The PR aforementioned is merged into dev now.

FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
```diff
@@ -703,19 +701,25 @@ extension OverrideProfilesConfig {
                                 label: "SMB IOB: "
                             )
 
-                            if ((aisf.smbDeliveryRatioMin ?? 0.5) as Decimal) != standard
-                                .smbDeliveryRatioMin || ((aisf.smbDeliveryRatioMax ?? 0.5) as Decimal) != standard
-                                .smbDeliveryRatioMax
+                            let smbDeliveryRatioMin = aisf.smbDeliveryRatioMin as? Decimal ?? standard.smbDeliveryRatioMin
```
yurique, line 704: if any of the three settings are changed - display the whole text (`SMB ratio: 0.3-0.7 range 5.6 mmol/L`)
if min=max - only display one value (`SMB ratio: 0.5`)

I wonder if `SMB ratio: 0.3-0.7 over 5.6 mmol/L` ("over" instead of "range") is better.

FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
```diff
@@ -689,10 +689,8 @@ extension OverrideProfilesConfig {
                             bool(bool: aisf.ketoProtect, setting: standard.ketoProtect, label: "Keto: ")
                             bool(bool: aisf.use_B30, setting: standard.use_B30, label: "B30: ")
 
-                            LazyHStack(spacing: 5) {
-                                decimal(decimal: aisf.autoisf_min, setting: standard.autoisf_min, label: "Min: ")
-                                decimal(decimal: aisf.autoisf_max, setting: standard.autoisf_max, label: "Max: ")
-                            }
+                            decimal(decimal: aisf.autoisf_min, setting: standard.autoisf_min, label: "Min: ")
+                            decimal(decimal: aisf.autoisf_max, setting: standard.autoisf_max, label: "Max: ")
```
yurique, line 693: it fills like it should be able to combine these two into one text with a range (`0.7-2.0`) - but I can't decide which label to use in that case

FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
```diff
@@ -689,10 +689,8 @@ extension OverrideProfilesConfig {
                             bool(bool: aisf.ketoProtect, setting: standard.ketoProtect, label: "Keto: ")
                             bool(bool: aisf.use_B30, setting: standard.use_B30, label: "B30: ")
 
-                            LazyHStack(spacing: 5) {
```
yurique, line 692: `LazyHStack` was adding large vertical padding for some reason

FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
Jon-b-m: No, because ”over” can mean several things. Better to be precise here.

FreeAPS/Sources/Modules/OverrideProfilesConfig/View/OverrideProfilesRootView.swift
Jon-b-m: No need to make more complex every body knows what min and max means. It’s clear and precise already imo.

yurique:
currently it looks like this:
image

yurique:
current before/after (still not perfect):

image

image

Jon-b-m:
OK. I think it looks better before

yurique:
How about this? (all tags are at the top, first line is white)

image

yurique:
with `Max IOB` on the first line 

image

yurique:
`Scheduled SMBs` as a tag as well

image

yurique:
some padding under the tags line

image

Jon-b-m:
It's a bit distracting having big AISF settings tags in first line. And I think it's a peculiar prioritising. The current dev/main has the sections in order of priority, not just by looks. 
The SMBs scheduled is also peculiar as a tag, because it's not exactly clear what the tag is meant to illustrate.  Does the red colour mean it's not scheduled? I therefore don't think that is very functional or helpful.

Jon-b-m:
I converted to draft, because it seems this PR is still in an idea/draft stage of development.

yurique:
> I find it a bit distracting having big AISF settings tags in first line.

I'd argue those settings are the most important when they are overridden - since the turn parts of the algorithm *completely* on or off:
* if I disable `accel` for the night - that's the biggest thing of the profile
* if I enable `B30` - that seems to be pretty important as well
* if autoISF is enabled/disabled in the override - it changes the algorithm completely
* disabled SMBs are important as well

> The SMBs scheduled is also peculiar as a tag, because it's not exactly clear what the tag is meant to illustrate

Maybe it could be a gray tag instead of red? This one is probably less important than "SMBs are off", but still important - for those who use it. 
I tried making it a tag so that it's in the same place as "SMBs are off", but it can go back to where it was. 
Or maybe it could have a separate line - with scheduled hours displayed?

---

I don't have a lot of real profiles on my phone, so it's harder for me to judge - would be nice to see what it looks with real profiles that people have. But I'd imagine most of the profiles will not have all the tags displayed - just one or two, and those will be important parts of the override.

---

On the other hand, I'm not really "married" to the idea of moving the tags to the top :) – I can move them back to where they were. But it's harder to prevent random un-pretty margins and gaps when tags can appear on multiple rows:

image

yurique:
reverted the tags to previous positions

before (current `dev`) / after (this branch):

image

image

yurique:
another direction to explore - everything is a tag:

image

image

Looks... maybe interesting, I guess? :) 

Something is adding random 1-2 pixel of extra space between some rows, and it's annoying. And I don't see why.

yurique:
current status:

image

yurique:
fixed text misalignment


image

yurique:
more padding between tags, formatted duration:

image

yurique:
a bit less padding between adjacent tags

image

yurique:
updated the description, the up-to-date screenshots are in there

Jon-b-m:
Thank you for the screenshots. And nice to finally have a description of what you’re actually intending with this PR…

Overall this suggestion still looks incomplete for the following reasons:

Profile 2,3,4 and 6 all look different from one another. There is no consistency and the spacing and padding isn’t the same. 

The added verbose time function isn’t needed in SwiftUI. There is a much cleaner ways of formatting the time. 

The ViewBuilder change here has no net effect (different UI).

yurique:
> Profile 2,3,4 and 6 all look different from one another.

Nothing is changed about this inconsistency in this PR - it's exactly the same as in current `dev` (the screenshots on the left are made in the `dev` branch).

> The ViewBuilder change here has no net effect (different UI).

👍 in that case it's not necessary, reverted

> There are cleaner ways of formatting the time.

replaced with `DateComponentsFormatter` (hours are always `h`, minutes are always `m`)

Jon-b-m:
To add to the list it's unfortunately also lacking localization. And in current form it can't be read and translated in Crowdin as you're using string  interpolation without  the necessary wrapping

Jon-b-m:
You're also mentioning padding and spacing in this PR, however hone of that is adressed?

yurique:
It will have to be a separate PR.

Jon-b-m:
I mentioned some days ago that I already had a local PR fixing everything requested with minimal code. But I wanted you to still have a change doing these changes yourself since you talked so much in private DMs and here about this PR. But it makes no sense merging something which isn’t fixing half of the issues you wanted to fix and asked me how to fix?

yurique:
At this point I really don't know what we're talking about. 

I wanted to make some adjustments. I implemented them. 
I did want to deal with padding issues as well. And I implemented the changes for that here initially, then iterated a few times with different suggestions/ideas.
Eventually, you told me I wasn't using something from `dev` which would make 80% of my changes unnecessary – I wasn't able to find what exactly I wasn't using, and I knew you already had a simpler fix for the padding issues – so it only made sense for me to remove the over-complicated changes in order to make the PR about the other things only, so we could stop going in circles about the padding. 

If the changes proposed here are wrong or not good enough - we can simply close this PR. No harm in doing that.

Jon-b-m:
Ok. I’ll close and push my old Pr.

PR #1195, opened 2025-03-10
Use the same names for Auto ISF strings in Settings and Override

luborjurena:


Jon-b-m:
Thanks. The translations with "... BG's" are not grammatically correct. I sugget use the existing strings with "...BGs" everywhere instead.

PR #1196, opened 2025-03-10
Add new Slovak translation strings

luborjurena:


Jon-b-m:
Great, but you need to add the source strings also in the en.lproj folder, otherwise nothing gets translated

luborjurena:
> Great, but you need to add the source strings also in the en.lproj folder, otherwise nothing gets translated

I checked that in multiple localizations, and if there is no string included in the language.lproj folder, then the string is loaded directly from the code. But maybe I'm wrong. In my dev, these translations works well.

Jon-b-m:
The benefit of having the string also in English is that we can push the new en.lproj to Crowdin project, where it will be auto generated for all languages and available for all translators in all languages

PR #1220, opened 2025-03-12
Add insulin types for translation.

luborjurena:


Jon-b-m:
I think these strings belongs in iAPS/Dependencies/LoopKit/LoopKit/Resources/ instead. However currently only Fiasp is localized here.

luborjurena:
I tested that in the simulator multiple times, but the localization files from iAPS/Dependencies/LoopKit/LoopKit/Resources/* are not loading for the insulins. I'll look at it again later.

Jon-b-m:
Some  Source strings seem to be missing. Only the Fiasp string exists (which is why you can see this particular string translated but not the others).

PR #1265, opened 2025-04-20
autoisf: report bg_isf more precisely (when < 1)

yurique:
it seems that sometimes bg_isf is calculated to be very close to 1 but still below it (like 0.999), so in the pill we report `bg: 1`, but the rest of the factors are not evaluated since the code shortcuts on `if (bg_ISF < 1)` - it gets pretty confusing


Jon-b-m:
Could you please describe the issue you’re having when ratio is 0.999?

PR #1314, opened 2025-05-17
change unit from sec/E to sec/U

luborjurena:
change unit from sec/E to sec/U to match pump display; translations are already using sec/U


Jon-b-m:
Great, thanks!

PR #1322, opened 2025-05-18
update string for "SMB Delivery Ratio BG Range description" and "iobTH"

luborjurena:


Jon-b-m:
Looks to be an extra ” in the added source string?

Jon-b-m:
Fixed it

PR #1341, opened 2025-05-21
Statically select Xcode 16.3 in GitHub Actions

luborjurena:
Manually sets the Xcode version because Xcode 16.0 is selected by default, which is not compatible and causes the build to fail.


Jon-b-m:
Did you test?

luborjurena:
Yes, I did and it's working.

PR #1368, opened 2025-05-28
show insulin activity and COB on main chart

yurique:
TODO:
* [x] colors - right now it's just white/black for dark/light themes
* [x] colors - regular blue/red looks better; not sure if we want some custom colors for this 
* [x] activity is actually represented by the "area under curve", so it might be better to show an "area chart" (with fill, not just stroke)
* [x] scaling - right now the activity is scaled to hard-coded range of `(minValue, maxValue) = (-0.005, 0.02)`, it's definitely wrong
* [x] better visual representation in general - right now there is no y-axis nor anything else to help interpret the activity data; we can only infer when it was/going to be relatively stronger/weaker
* [ ] saving the data - it's probably better to do this outside of `OpenAPS` class 
* [ ] might be good to have this on the Activity widget as well
* [ ] other tweaks and code-style improvements

<details>
  <summary>A bigger/broader topic:</summary> 
* we are currently showing the data from *before* the suggestion was enacted (which means IOB and predictions are "outdated" and are not reflecting the current state of the "system");
* it has always been like this, but it would be really nice to run oref0 one more time *after* the suggestion is enacted (or any treatment, not just suggested) to show the up-to-date information.

Resolution: this will be a separate PR/feature.
</details>

### Current screenshots: 

<details>
  <summary>previous screenshots "v4"</summary> 
  
  
</details>

<details>
  <summary>previous screenshots "v3"</summary> 
  
  
</details>

<details>
  <summary>previous screenshots "v2"</summary> 
  
  
</details>

<details>
  <summary>previous screenshots "v1"</summary> 
  
  
</details>


FreeAPS/Sources/Models/IOBTick0.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+// "0" for oref0
+struct IOBTick0: JSON {
```
yurique, line 4: this is a stripped down version of `IOBEntry` - that one has data we don't need and formats for dates are mixed (sometimes it's a string, sometimes it's a number of seconds - the JSON decoding was failing)

Jon-b-m:
To have the view updated when enacted is possible with small update. I can push this separately as it would be a good standalone feature

Jon-b-m:
The insulin activity is a good idea.

yurique:
### previous commit summary:

* replaced .concurrentMap with just .map in a few functions (including existing functions) in the chart view - it seems like messing with UnsafeMutableRawPointer just to go through a couple hundred numbers is an overkill

* changed the core data entity fields to be non-optional (I missed it initially, because I didn't know about the inspector) (xcode still generates the fields as optional, so it's handled after fetching)

* removed all unwraps - no more nils should be possible anymore, because of 👇

#### in the new CoreDataStorage.fetchInsulinData function: 

* changed it to return an array of IOBTick0 (the same struct that is used when storing) instead of the code-generated InsulinActivity - to avoid passing NSManagedObject objects around threads/queues (that's my only suspect for those nil values in the entities), and also to ensure there are no nils (fields in IOBTick0 are non-optional)

* filter out results where any of the fields is nil for whatever reason (just .compactMap)

* changed sorting to ascending (it probably doesn't matter, but it makes more sense)

#### in the new CoreDataStorage.saveInsulinData function:

* when deleting previous "future" entries - delete up to 60 seconds before the new entries (to fix weird "steps" in the activity chart around loop timestamps)

Jon-b-m:
How would it look if you aligned the insulin amount label to the left instead? 

I also think we should make the display of insulin activity optional, as new settings in UI/UX settings.

yurique:
with Insulin Activity toggle in settings:

yurique:
> Yes, it does a look a bit weird…

I pushed the changes putting it on the left, but I can revert :)

yurique:
updated the screenshots in the description (with blue background)

PR #1373, opened 2025-05-28
Localize hour buttons

luborjurena:


luborjurena:
Yes, partially. But currently, it is not possible to use the correct format for translation. Only the translation for “hours” is used from the language file; translations for whole strings like “3 hours”, “6 hours”, etc., are currently ignored.

image

Jon-b-m:
Ok

Jon-b-m:
Could you please also add ”24 hours” to translate all picker options? We also need a small change in the picker itself, some small localization edit.

luborjurena:
"24 hours" is already in the translation file

Jon-b-m:
> "24 hours" is already in the translation file

OK! 👍 

But for Slovak we actually only need to add 3 and 6 hours for translations? Are other translations needed for other Slavic languages?

PR #1392, opened 2025-06-04
Refactored timeSetting menu to use localized strings for each hour value

luborjurena:
Translation and label of all hour strings is now handled with the language file

Screenshot 2025-06-04 at 09 32 27
Screenshot 2025-06-04 at 09 32 21
Screenshot 2025-06-04 at 09 32 18


Jon-b-m:
Great, thanks.

PR #1395, opened 2025-06-04
disable autoisf completely when it's disabled by middleware

yurique:
Currently, if autoisf is disabled in middleware, the autoisf code is still executed (unless there is ALSO an override active), which causes the `SMB Delivery Ratio` to reset to default `0.5` (ignoring the profile settings).

This PR prevents autoisf from executing when `profile.iaps.autoisf` is set to `false`, regardless if it's disabled by middleware or an override.


Jon-b-m:
Thanks!

PR #1414, opened 2025-06-09
basal chart rendering - make more resilient to corrupted data

yurique:
A problem reported in Discord turned out to be caused by the following corruption of pump history:
* a `TempBasalDuration` history record without the corresponding `TempBasal` record
* this causes all subsequent TBRs to be missing from the main chart

The reason:
* the `data.tempBasals.chunks` splits the history in chunks of two, expecting each pair to be exactly a `TempBasal` record followed by `TempBasalDuration`, every other pair is ignored
* an orphan `TempBasalDuration` record breaks this expectation completely - the order in the pairs becomes reverted 

The fix:
* to use the `window` function instead of `chunks`, making the procedure resilient to this type of corruption

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/Chart/MainChartView.swift
+++ b/FreeAPS/Sources/Modules/Home/View/Chart/MainChartView.swift
@@ -917,7 +917,7 @@ extension MainChartView {
                 fullSize: fullSize,
                 autotuned: false
             )
-            let tempBasalPoints = firstRegularBasalPoints + data.tempBasals.chunks(ofCount: 2).map { chunk -> [CGPoint] in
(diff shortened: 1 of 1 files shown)


yurique:
> OK! I assume this has been tested?

Yes, I tested this in the simulator using the corrupted pump history file from Discord - it works and it handles the missing entry gracefully. 

The pump history also contained a mismatched pump suspension - there was a "resume" without the corresponding "suspend" - and this is handled gracefully already because for suspensions we're already using the `.windows` function:
```swift
var rects = data.suspensions.windows(ofCount: 2).map ... 
```

I ran it on my phone as well - everything works and looks fine.

yurique:
Here's a fix for the underlying problem (hopefully): https://github.com/Artificial-Pancreas/iAPS/pull/1418

PR #1348, opened 2025-05-24
feat: Replace duplicated bolus Entries & sync DanaKit

bastiaanv:
Changelog:
- Sync DanaKit with latest dev
- Replace duplicated bolus entries in iAPS (entries are matched on `startDate`)

This last change allows DanaKit to preemptively record a bolus entry to be replaced with the actual bolus entry later.
This fixes a lot of missing bolus' for DanaKit, because of connection loses.
This method is also used in Trio (dev) & Loop, but wasnt implemented here because of this missing functionality.

> [!WARNING]
> This is not tested with OmniBLE or any other pumpManager and requires additional testing before merging

Code used to test the changes in DanaKit:
```swift
public func test() {
        let doseEntry = UnfinalizedDose(
            units: 5,
            duration: .minutes(5),
            activationType: .manualNoRecommendation,
            insulinType: .fiasp
        )
        doseEntry.deliveredUnits = 5
        
        self.pumpDelegate.notify { delegate in
            let dose = doseEntry.toDoseEntry()
            delegate?.pumpManager(
                self,
                hasNewPumpEvents: [
                    NewPumpEvent
                        .bolus(
                            dose: dose,
                            units: dose.programmedUnits,
                            date: dose.startDate
                        )
                ],
                lastReconciliation: Date.now,
                completion: { _ in }
            )
        }
        
        // Wait for 2sec before recording a duplicated bolus
        DispatchQueue.main.asyncAfter(deadline: .now() + 2.0) {
            doseEntry.deliveredUnits = 7
            self.pumpDelegate.notify { delegate in
                let dose = doseEntry.toDoseEntry()
                delegate?.pumpManager(
                    self,
                    hasNewPumpEvents: [
                        NewPumpEvent
                            .bolus(
                                dose: dose,
                                units: dose.programmedUnits,
                                date: dose.startDate
                            )
                    ],
                    lastReconciliation: Date.now,
                    completion: { _ in }
                )
            }
        }
    }
```

PR #1418, opened 2025-06-10
pump history storage - fix bolus de-duplication [fixes #1416]

### LINKED ISSUE #1416: Pump history corruption (with pump simulator)
related to https://github.com/Artificial-Pancreas/iAPS/pull/1414

The pump history json has these two entries:

```json
[
  {
    "timestamp" : "2025-06-09T08:53:15.423Z",
    "_type" : "TempBasalDuration",
    "id" : "cec0de001bf1aa7c06ba9fabdb5ff54a",
    "duration (min)" : 30
  },
  {
    "amount" : 0.3,
    "id" : "_cec0de001bf1aa7c06ba9fabdb5ff54a",
    "duration" : 0,
    "timestamp" : "2025-06-09T08:53:15.423Z",
    "_type" : "Bolus",
    "isExternal" : false,
    "isSMB" : true
  }
]
```

Notice how the `Bolus` event has:
* `id` with underscore `_`
* the rest of the ID matches the `TempBasalDuration` id

Which means:
* the TBR pair (rate + duration) was created first, with IDs `cec0de001bf1aa7c06ba9fabdb5ff54a` and `_cec0de001bf1aa7c06ba9fabdb5ff54a`
* after that, a bolus event was stored and the de-duplication logic erroneous picked up the `TempBasal` record and replaced it (that's the only way a bolus record would have an underscore in the ID):
```swift
    func storePumpEvents(_ events: [NewPumpEvent]) {
    // ...
                case .bolus:
                    // ...
                    if let duplicatedEvent = storedEvents
                        .first(where: { x in Int(x.timestamp.timeIntervalSince1970) == Int(event.date.timeIntervalSince1970) })
                    {
                       // ... replace the "duplicate" with this bolus event
                        return [PumpHistoryEvent( 
                            id: duplicatedEvent.id,
                            type: .bolus,
                            timestamp: duplicatedEvent.timestamp,
                            // ...
                        )]
                    }
```

The reason it picked up the TBR rate as a duplicate because the timestamp happened to be exactly the same (because of how simulator works, I suppose) and the de-duplication only looks at timestamps.

yurique:
Fixes #1416

The issue was introduced in this commit (May 2025) - https://github.com/Artificial-Pancreas/iAPS/commit/6279502fbde30523c0a0d535b039522660e43745 (part of this PR: https://github.com/Artificial-Pancreas/iAPS/pull/1348)

The de-duplication logic looks for an existing bolus entry at the exact same timestamp and, if found, replaces it instead of creating a new record. 

The problem is the code is not looking at the type of the existing record and can replace other events instead of boluses - like TBR rate/duration when the timestamp matches exactly. 

I'm not sure if this is possible when using a real pump, but it does happen with the pump simulator (as described in the issue and on Discord).

This PR adds a check such that we only look at "bolus" events when de-duplicating (so the timestamp has to match and the type has to be "bolus").

(related to https://github.com/Artificial-Pancreas/iAPS/pull/1414)

### DIFF
--- a/FreeAPS/Sources/APS/Storage/PumpHistoryStorage.swift
+++ b/FreeAPS/Sources/APS/Storage/PumpHistoryStorage.swift
@@ -50,7 +50,8 @@ final class BasePumpHistoryStorage: PumpHistoryStorage, Injectable {
 
                     let minutes = Int((dose.endDate - dose.startDate).timeInterval / 60)
                     if let duplicatedEvent = storedEvents
-                        .first(where: { x in Int(x.timestamp.timeIntervalSince1970) == Int(event.date.timeIntervalSince1970) })
+                        .first(where: { x in
+                            Int(x.timestamp.timeIntervalSince1970) == Int(event.date.timeIntervalSince1970) && x.type == .bolus })
                     {
                         return [PumpHistoryEvent(
                             id: duplicatedEvent.id,


yurique:
looks like this might be affecting some real pumps as well: https://discord.com/channels/1120154740857245808/1124371781793349662/1381186986257416253

PR #1443, opened 2025-06-22
Add Hide Settings Toggle

scrappy:
This adds a Hide Settings toggle switch to iOS Settings for the app, and turns off all Settings *except* for allowing to Open/Close Loop

When toggled on, a message is provided to remind folks where to re-enable it

The idea isn't to make it impossible to make changes, just to add an extra step to prevent someone from accidentally making changes ( ie. someone nosy looking around )


scrappy:
done

PR #1484, opened 2025-07-13
fix(MedtrumKit): prime progressbar is now updating

bastiaanv:
Small MedtrumKit fix, so that the prime progressbar does update during priming


bastiaanv:
Tagging @Jon-b-m

Jon-b-m:
Thanks!

Issue #1476, opened 2025-07-08
Issues with bolus after cannula refill from iAPS


Dia-muff:
After running the function Settings -> Pump -> Configuration -> Reservoir/cannula refill -> Cannula only -> Save, cannula is filled, but any bolus sent by iAPS to pump is delivered by the pump but not shown on the graph and not registered in history. Additionally, these boluses are not counted towards insulin on board, and therefore not taken into consideration in next cycle.
In the attached log file, last bolus (before manually triggering cannula refill) was delivered at 2025-07-08T11:24:30+0200 timestamp. This bolus was shown on graph and counted towards insulin onboard. Until 16:34 (manual iPhone restart, as closing/opening iAPS did not help), no bolus was shown on the graph and counted towards insulin on board, while the pump was delivering the boluses.
I attach log from the iAPS and photo of the app.

![Image](https://github.com/user-attachments/assets/061ac47f-ddb3-4987-9534-8db58bba71a0)
![Image](https://github.com/user-attachments/assets/6e4712a3-ecf1-487b-a778-d002d326a9bc)
![Image](https://github.com/user-attachments/assets/8bb8012c-61b4-46a5-9787-d89f8abd82b4)
![Image](https://github.com/user-attachments/assets/cb15c49d-3eae-4c08-b983-50ae7e11fd68)
![Image](https://github.com/user-attachments/assets/6a598518-df52-48df-8350-72dfe36cfdfd)
![Image](https://github.com/user-attachments/assets/52706e1a-bc5a-471b-924a-37b7c2acd9d7)
![Image](https://github.com/user-attachments/assets/8d8756b9-62d9-4190-bcf7-07c51e68873e)
![Image](https://github.com/user-attachments/assets/555eb096-60d8-441a-9996-82f6585b6b7f)
![Image](https://github.com/user-attachments/assets/33db22c9-a256-421e-82b0-667cc529d7f7)
![Image](https://github.com/user-attachments/assets/b32ec459-7527-4ba2-947e-96333af2c2f9)
![Image](https://github.com/user-attachments/assets/f085757e-3015-4f99-acea-ef83aea10bc0)
![Image](https://github.com/user-attachments/assets/22f921cc-a3c7-4435-9aa1-5058a9c17fb9)
![Image](https://github.com/user-attachments/assets/245885d1-413f-46f5-9ac5-39cfd0f643b2)

[log.txt](https://github.com/user-attachments/files/21129650/log.txt)

[...]

PR #1485, opened 2025-07-13
fix(DanaKit): Reset priming mode after completion

bastiaanv:
DanaKit has a feature, where you can refill the reservoir and/or replace the cannula of your pump, without using the pump's UI. In order to prevent the priming to be stored as a bolus, DanaKit can be put into priming-mode. Due to a coding error, DanaKit did not switch back to a normal mode after a successful prime causing the issue as describe in #1476


Jon-b-m:
Perfect 👌

PR #1512, opened 2025-07-28
Export BUILD_GROUP variable from Github Settings -> Fastlane

scrappy:
I had written the code months ago to allow setting COPYRIGHT_NOTICE via a BUILD_GROUP variable in Github Actions, but failed to export that variable so that Fastlane could use of it.

This corrects that, and adds an output of ConfigOverride.xcconfig step so that its easier to debug going forward


Jon-b-m:
OK. Is it tested?

PR #1546, opened 2025-08-07
less chart extra width when predictions are hidden

yurique:
https://github.com/user-attachments/assets/f4ffd411-ce3c-442d-94ca-f679f5003827


Jon-b-m:
But perhaps we need to see more than 60 minutes into the future, for instance when planning for an exercise? Perhaps 90 minutes would suffice?

yurique:
right, makes sense 👍 
Will push an update in a few minutes.

yurique:
latest:
image

PR #1556, opened 2025-08-11
Support for newer EU Libre 2+ sensors

Jon-b-m:
Not tested yet, might be more changes needed...

Commits from LoopKit (dabear)

### DIFF
--- a/Dependencies/LibreTransmitter/Sources/LibreTransmitter/Bluetooth/LibreTransmitterMetadata.swift
+++ b/Dependencies/LibreTransmitter/Sources/LibreTransmitter/Bluetooth/LibreTransmitterMetadata.swift
@@ -1,4 +1,4 @@
-//
+///
 //  MiaoMiao.swift
 //  LibreMonitor
 //
@@ -85,6 +85,7 @@ public enum SensorType: String, CustomStringConvertible {
     case libreUS14dayE6 = "E6"
     case libreProH = "70"
     case libre2Plus = "C6"
+    case libre2PlusNew = "F7"
 
     public var description: String {
         switch self {
@@ -92,7 +93,7 @@ public enum SensorType: String, CustomStringConvertible {
             return "Libre 1"
         case .libre1A2:
             return "Libre 1 A2"
-            case .libre2, .libre2C5, .libre2Plus:
+            case .libre2, .libre2C5, .libre2Plus, .libre2PlusNew:
             return "Libre 2"
         case .libreUS14day, .libreUS14dayE6:
             return "Libre US"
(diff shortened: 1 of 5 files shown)


aurelien76:
Hello, I've been using libre 2+ EU for a while now and I can't connect them directly to IAPS. I also tried with your version and the latest 7.0.1, without success, but they work in TRIO or XDRIP for example. Is this currently being developed for IAPS?

Jon-b-m:
@aurelien76 thanks for feedback. I’ll update the PR. It’s a bit tricky when you don’t have the actual sensors at hand.

aurelien76:
@Jon-b-m I'm trying this coconut when I get home from work. 
Should I test it with the L2+new version that's online?

Thank you very much for your work!!

Jon-b-m:
Yes, you don’t need to change your sensor if already using Libre2plus

aurelien76:
Hum unfortunately it's the same, when you want to add it and click on free direct, it opens the menu with device but it also reopens another free direct/Bluetooth selection menu, the appanage works and the initial pairing menu disappears but we go around in circles on the menu that reopened with direct/bluetooth

Jon-b-m:
> Hum unfortunately it's the same, when you want to add it and click on free direct, it opens the menu with device but it also reopens another free direct/Bluetooth selection menu, the appanage works and the initial pairing menu disappears but we go around in circles on the menu that reopened with direct/bluetooth

Thanks för testing. Could you please do one thing for me? Connect iPhone to your Mac, run iAPS from Xcode and try a few times to connect your Libre 2 plus, while still connected.  After that copy everything from the console in a txt file and attach here. Thanks in advance.

aurelien76:
it does that on the phone, I recorded the screen


https://github.com/user-attachments/assets/ce30c15f-f1db-4981-b29d-5fb5e2442a6d

gui-dos:
The culprits are two typos: `F7` instead of `7F` in https://github.com/Artificial-Pancreas/iAPS/blob/c183bd257041b700d2b6bdec60ec70d060979f14/Dependencies/LibreTransmitter/Sources/LibreTransmitter/Bluetooth/LibreTransmitterMetadata.swift#L88 and https://github.com/Artificial-Pancreas/iAPS/blob/c183bd257041b700d2b6bdec60ec70d060979f14/Dependencies/LibreTransmitter/Sources/LibreTransmitter/Bluetooth/LibreTransmitterMetadata.swift#L112

sandrotsurkava:
Hi all, 

Chiming here with a similar issue, I have 3 Libre 2 plus sensors, so for the next 1.5 months I can also test the version. Currently my logs from main branch show similar to @aurelien76's initial logs.

If testing and logs are needed, I can provide them per instructions.

Jon-b-m:
> This project should consider updating the code to use the latest libretransmitter sub module directly, rather than relaying on monkey patching the forked code directly In the Iaps peoject

PR would be most welcome.

dabear:
I understand the Trio guys have done that, maybe have a look there.

aurelien76:
Hello, I tried with TRIO to see if the connection would work, but I've never used it with that. So I installed it and tried to connect, the sensor seems to scan correctly, it pairs and is connected. I can see the serial numbers, remaining time, etc., but I don't see anything on the main screen for blood glucose levels. This may be normal use of TRIO, but it doesn't seem to be showing up (even though it connects, pairs, etc.).
It works too.

sandrotsurkava:
Just to add context, I switched to Zukka app (an appstore mirror of xDrip4iOS) and it works well for L2+ sensor.

Jon-b-m:
Closing. Will create new PR with updated LoopKit modules.

PR #1567, opened 2025-08-18
simplifying oref0 integration

yurique:
### a single `.js` bundle

The webpack build now builds a single `js` bundle which includes all the upstream oref0 functions, as well as all the custom scripts from `prepare/*`.

* [ ] scripts from `prepare/*` now live inside `scripts/src` - I didn't remove the existing ones yet (we can make webpack reference `prepare/*`, but it will be easier to forget to rebuild the bundle after making changes there)

The bundle exposes a single entry point: `iaps.invoke(functionName, iapsInput) { ... }`.

The webview only needs to load this one js bundle, and invoke the functions using the single entry point:
```js
iaps.invoke(\"\(function)\", input)
```
* where `function` is the name of the function supported by the entry point, like "iob", "determine_basal", etc
* and input is a (typed) JSON object containing all the inputs for the function

### webview simplification

The "post-message" mechanism is not needed to return results from the evaluated js-functions. So I simplified this as well.
(there is a workaround for simulators because of a "bug": https://developer.apple.com/forums/thread/779012)

Middleware invocation is also simplified - the body of the user's middleware function is passed into js as a string and evaluated there using `eval`.

In overall, `WebViewScriptExecutor` is pretty small and simple now.

### types

All data going in and out javascript is now typed. Added missing types, as well as per-function "input" structs.

The next step would be to rewrite all  custom `prepare/*` scripts in Swift (except autoisf).

There were multiple places in the code where things like `Decimal` where parsed from strings using decoding from the JSON helper. Replaced those invocations with the regular `Decimal(string: ...)` constructors.

### error handling

Errors from webview were silently ignored because in OpenAPS, all functions returned `Future<..., Never>`. So in case of a failure, the future just never completed and the app remained in "spinning" mode.

Changed those functions to return `Future<..., Error>` and added the corresponding handling in `APSManager`.

### autotune simplification

We are calling `autotune-prep` only to get the result and pass it into `autotune-core`. Now it's combined into one js function that does both steps.

---

This will need more testing, obviously. I haven't even tried running it on the phone yet. But it seems to work just fine in the simulator so far.


yurique:
this PR got completely outdated, and the same simplification is now implemented in the `alpha` branch

PR #1571, opened 2025-08-22
New defaults

scrappy:


Jon-b-m:
Looks like you're missing an update in FreeAPS/Sources/Models/FreeAPSSettings.swift also.

PR #1577, opened 2025-08-27
nightscout remote push notifications

yurique:
Remote notifications for nightscout-as-CGM.

Requires a counterpart backend server (https://github.com/yurique/iaps-ns-apns), which subscribes to updates from Nightscout and sends remote notifications via APNs.


FreeAPS/Sources/Services/Network/NetworkService.swift
yurique, line 1: here - just adding the `async` versions, the old method remains as is

PR #1582, opened 2025-08-31
updating to plugin-architecture with dependencies as git-submodules (the LoopWorkspace way)

yurique:
## Plugins 

This PR introduces the "plugin architecture" following the Loop Workspace pattern.

Each CGM and pump are added to the workspace as "plugins" and discovered dynamically in runtime.

<details>
  <summary>How plugins work</summary>
  
  Each plugin has an additional "Plugin" target, for example `G7SensorPlugin`.
  
  * this target has the actual implementation targets as dependencies;
  * provides a "Principal class", for example `G7SensorPlugin.swift`
  * this principal class is specified in the plugin's `Info.plist` under `Principal class` entry
  * this pricnical class implements the corresponding protocol (`PumpManagerUIPlugin`, `CGMManagerUIPlugin`, etc)
  * additionally, the `Info.plist` has entries for the main app to discover and categorize the plugins:
    * `com.loopkit.Loop.CGMManagerIdentifier` and `com.loopkit.Loop.CGMManagerDisplayName` for CGM plugins
    * `com.loopkit.Loop.PumpManagerIdentifier` and `com.loopkit.Loop.PumpManagerDisplayName` for pump plugins
  * the "plugin" targets also have `Wrapper Extension` set to `loopplugin` in the Build Settings (instead of the default `framework`)

  The main app build steps include an extra phase - `Install Plugins`, which invokes the `${SRCROOT}/FreeAPS/Scripts/copy-plugins.sh` script.
  This script looks for `*.loopplugin` bundles and post-processes them.

  At start up, the app scans the available bundles, checks the `Info.plist`, and decides which bundles are plugins, and which kind of plugin they are.

  After that, we have two lists: `availablePumpManagers` and `availableCGMManagers` which are used by the `DeviceDataManager`.
</details>

A new scheme is introduced - `FreeAPSWorkspace` (`Product` / `Scheme` / `Edit scheme...`):
* the main app is added as a target
* all the plugins are added as targets
* the main app no longer depends directly on plugins

Build order is set to `Manual Order` (as opposed to default `Dependency Order`) in the scheme settings. It should work with `Dependency Order` as well, and I did have successful builds like that - but eventually xcode broke and started failing with random resolution errors. This is probably worth to investigate further.

## Git sumbodules

All plugins are now git-submodules and it can be tricky not to mess up.

<details>
  <summary>How to manage submodules</summary>

  I used the Loop Workspace `dev` branch as the reference.
  To see which commits Loop Workspace is referencing:
  ```shell
  LoopWorkspace $ git submodule status
  # output:
  # 98f9759b17e8716c65257c24a652564ead3c5183 AmplitudeService (heads/dev)
  # 68ab20c65992a4db94bab5df73e5b9f9b615e4b0 CGMBLEKit (v3.1-169-g68ab20c)
  # a8e4208572738d310db2a1991195c19c19e33c15 DanaKit (remotes/origin/trio_for_0.2.x-33-ga8e4208)
  # a159c7fca083147a66a41783d38cfc65fac3a9b0 G7SensorKit (remotes/origin/HEAD)
  #...
  ```

  We see the commits to which the submodules are "pinned" as well as the corresponding remote branches/tags.

  To see the remote URLs for the submodules:
  ```shell
  LoopWorkspace $ git config -f .gitmodules --get-regexp '^submodule\..*\.url'
  # output:
  # submodule.Loop.url https://github.com/LoopKit/Loop.git
  # submodule.LoopKit.url https://github.com/LoopKit/LoopKit.git
  # submodule.CGMBLEKit.url https://github.com/LoopKit/CGMBLEKit.git
  # submodule.dexcom-share-client-swift.url https://github.com/LoopKit/dexcom-share-client-swift.git  
  # ...
  ```

  Now we can add the submodules to the FreeAPS repo:
  ```shell
  git submodule add https://github.com/LoopKit/OmniKit.git OmniKit
  # repeat for all
  ```

  And then we make sure to pin the submodule to the same commit as in Loop Workspace:
  ```shell
  git -C OmniKit checkout c2ec2fd29a31046ff784ac0c11a9e0b36bc14c2a
  # repeat for all
  ```

</details>

This also means that when cloning/pulling the repository we need to pull the submodules as well.

```shell
# clone
git clone --recurse-submodules <repo-url>
# pull
git pull --no-recurse-submodules
git submodule update --init --recursive --checkout
```

To make `git pull` not mess up the submodules (by default it will pull them to the latest commit in the corresponding branches) without specifying `--no-recurse-submodules` set this git config option:
```shell
# in the iAPS repository directory:
git config submodule.recurse false
```

* [ ] The fastlane actions will need to be updated.

## CGM refactoring 

All the CGM-specific `GlucoseSource`s are removed (as well as the `GlucoseSource` protocol itself). All interaction with CGMs is handled by the `DeviceDataManager`. The `FetchGlucoseManager` is removed. A new class is introduced - `BloodGlucoseManager`. Its only responsibility now is to store new blood glucose readings (+ push to Nightscout and HealthKit). Whenever the `DeviceDataManager` gets new blood glucose readings - it invokes the `storeNewBloodGlucose` function from this class.

Instead of custom "Glucose Source" for Nightscout, the Nightscout Remote CGM plugin is now added.

The `Use local glucose source` and `Local glucose port` settings are removed. Instead, the user should enter the local address (`http://127.0.0.1:$PORT`) in the Nightscout Remote CGM settings.

## Pump refactoring

Not much has changed with regards to pump management. The pump managers are now also discovered from plugins, but the integration was already "generic", without custom code per pump.

Some code was changed/introduced (following the patterns from Loop) related heartbeat-from-pump. This part needs verification/testing.

<details>
  <summary>Pump as CGM</summary>
  When the configured pump manager is also a CGM manager (minimed) - it will show up in the list of available CGMs.
  When it is selected as the CGM, the UI will say that the pump is used as a CGM and a button to stop using the pump as CGM will be presented.

</details>

* [ ] `TODO:` tweak the text/message in the UI for the "pump is cgm" case

## Loop triggering

The code-paths from heartbeat/glucose reading to loop have been refactored and simplified.
Now the `DeviceDataManager` is fully responsible for triggering the loops with two paths:

* `CGM new reading` -> `processReceivedBloodGlucose` -> `store new blood glucose` -> `update pump data` -> `recommend loop`
* `heartbeat` -> `read from cgm` -> `processReceivedBloodGlucose` -> `store new blood glucose` -> `update pump data` -> `recommend loop`

## Shared App Group Source refactoring (xDrip, etc)

The AppGroupSource has been converted into a CGM plugin, so it integrates into the standard flow in the `DeviceDataManager` and CGM-related UI.

## Non-standard CGM/Pump handling

We're extracting some extra information from the CGM and pump managers:
* sensor activation date (for Dexcom and Libre sensors)
* POD expiration date (for DASH pods)

Unfortunately, I'm not seeing any standard API to get this information in the LoopKit protocols.

We already had custom implementations for different devices, done by casting CGM managers and pump managers to concrete implementation. 

This remains unchanged, but all manager-specific code was moved into a single class - `KnownPlugins`.

## Blood glucose handling

* all glucose received from the CGM is stored and displayed
* we no longer read blood glucose from HealthKit
* when running a loop, the glucose records are filtered, forcing at-least 5-minute intervals 
* (a new developer setting to allow sending glucose with 1-minute intervals to oref, default: off)
* (a new developer setting to allow running a loop with 1-minute intervals, default: off)

## View state model injection

The dependency injection for view state model is changed a little.
We used to have this pattern where each view received a `Resolver`, and had a `configureView` (from the base view class) that would inject the dependencies into the state model. 

For this to work, each view had to have this: `.onAppear(perform: configureView)`.

This worked more or less well, but this meant that the state model was not fully initialized during the first render of the view - because `.onAppear` hasn't happened yet.

Now this machinery is removed, and the base state model now requires a `Resolver` in its `init`. This added a little bit of boilerplate - instead of `.onAppear`, the views now have to initialize the state model in the constructor:

```swift
        @StateObject var state: StateModel

        init(resolver: Resolver) {
            self.resolver = resolver
            _state = StateObject(wrappedValue: StateModel(resolver: resolver))
        }
```

But it is less fragile.

I'm open to suggestions to change/improve this. We might also consider using the `@EnvironmentObject` somehow, instead of `Resolver` (in the views). But this is not really relevant to this PR.

## Upload readings setting

The `Upload reading` toggle and setting is removed. We are using what the CGM plugin tells us.
Previously, we stored this in the settings and synchronized it with the CGM manager.

I'm open to revert/change this.

## Calendar settings

Calendar settings are moved out of the CGM screen into a separate screen.

## BG tap

Tapping on BG indicator on home screen opens the CGM settings view for real BT CGMs. Otherwise - opens app URL. We can revert this back to opening the first-party app, but this will work only for Dexcom CGMs.

## Other changes

I might have missed a few pieces - will update as I recall.

The PR is not done yet, so I'll be adding more sections. But nothing big is coming: need to refactor a few places that I just commented out for now, and more cleanup.


Jon-b-m:
SAGE looks to be in seconds now?

image

Jon-b-m:
To do: The bolus recommendation in Bolus View is not working anymore. Instead of a new suggestion errors are returned, for instance last loop to recent or in bolus View.

yurique:
> SAGE looks to be in seconds now?

yeah, there was a commit with this issue - it should be good now?.. or are you seeing this with the latest commit?

> To do: The bolus recommendation in Bolus View is not working anymore

this is next on my list 👍

Jon-b-m:
> yeah, there was a commit with this issue - it should be good now?.. or are you seeing this with the latest commit?

Let med test again (tomorrow) 👍

yurique:
## BG tap

Tapping on BG indicator on home screen opens the CGM settings view for real BT CGMs. Otherwise - opens app URL. We can revert this back to opening the first-party app, but this will work only for Dexcom CGMs.

yurique:
> The bolus recommendation in Bolus View is not working anymore

fixed
it was because of the view state model initialization changes - the state model for bolus calculator was "subscribing" too soon (before the views were setting all the needed vars)

yurique:
updated description:

---


```shell
# clone
git clone --recurse-submodules <repo-url>
# pull
git pull --no-recurse-submodules
git submodule update --init --recursive --checkout
```

To make `git pull` not mess up the submodules (by default it will pull them to the latest commit in the corresponding branches) without specifying `--no-recurse-submodules` set this git config option:
```shell
# in the iAPS repository directory:
git config submodule.recurse false
```

Jon-b-m:
> I was just trying to unify it a little bit.. for Dexcom - it was the 1st party app, for Libre - it was the built-in settings screen. Plus, in the Dexcom settings view there is a button to open the native app.

I see. I'm not sure. If we can open the open within this next View, then it's OK.

Jon-b-m:
> In NS settings there's no button to open it in the browser, but we can easily have a special case for NS.

Yes, I think this would be nice

yurique:
The `Use local glucose source` and `Local glucose port` settings are removed. Instead, the user should enter the local address (`http://127.0.0.1:$PORT`) in the Nightscout Remove CGM settings.

---

The AppGroupSource has been converted into a CGM plugin, so it integrates into the standard flow in the `DeviceDataManager` and CGM-related UI.

yurique:
## Blood glucose handling

* all glucose received from the CGM is stored and displayed
* we no longer read blood glucose from HealthKit
* when running a loop, the glucose records are filtered, forcing at-least 5-minute intervals 
* (a new developer setting to allow sending glucose with 1-minute intervals to oref, default: off)
* (a new developer setting to allow running a loop with 1-minute intervals, default: off)

PR #1585, opened 2025-09-06
Showing both Override and Temp target

i3bdulr7man:
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 16 41 58
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 16 42 06
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 16 42 13

### DIFF
--- a/FreeAPS/Sources/Models/FreeAPSSettings.swift
+++ b/FreeAPS/Sources/Models/FreeAPSSettings.swift
@@ -50,6 +50,8 @@ struct FreeAPSSettings: JSON, Equatable {
     var useCalc: Bool = true
     var fattyMeals: Bool = false
     var fattyMealFactor: Decimal = 0.7
+    var fastMeals: Bool = false
+    var fastMealFactor: Decimal = 1
     var displayPredictions: Bool = true
     var useLiveActivity: Bool = false
     var liveActivityChart = false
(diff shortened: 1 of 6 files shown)


charlesstyves:
Great work, thanks!

UI/UX opinion change with each person. May I suggest this : 
- Using the "  |   " separator is a great idea.
- Must user would probably prefer to keep the one line statut bar, to achive this, I suggest these compromises : 
-> Removing the two icons. They give a really nice look, but they took too much space... I guess that user should understand by themself that the right text refer to the right icon (override) and the left text refer to the left icon (temps target). Considering than the menu icon is highlighted, user should understand by himself if only one is activated, which one is it. Icons could also look too much for user using emoji as description.
-> Limit text showing by a certain lenght. If it exceed, user must consider using small description or emoji instand of text for his override or temp target (if he want to get full showing).

Thanks again to have give your time on this request!

i3bdulr7man:
@charlesstyves I agree that the icons could become confusing, especially for users who name their overrides or temporary targets with emojis. However, I still believe it’s important to visually distinguish between the two. One simple solution could be to differentiate them by color.

Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 22 19 49
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 22 19 58
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 22 20 07
Simulator Screenshot - iPhone 16 Pro - 2025-09-06 at 22 20 11

PR #1607, opened 2025-09-20
more cleanup and refactoring

yurique:
* AlertHistoryStorage minor refactoring
* DeviceDataManager - cleanup
* moved more pump-specific code into KnownPlugins
* simplified alert acknowledgement (don't rely on completion callback)
* delivery uncertainty message in pump config view

---
* the whole `PersistedAlertStore` protocol - it's supposed to be used by plugins to query the "storage" about existing alerts, but none of the methods are used in any of the pumps, so we just don't do anything there (same as before the `plugins` branch)
* detectedSystemTimeOffset - we're supposed to report the offset to the pump manager, and loop does have an implementation - but none of the pump managers is using it, we are returning `0` (same as before `plugins`)
* removed the whole `reportPluginInitializationComplete` thing (which I initially "copied" from Loop) - Loop notifies plugins when other plugins activate, but this has no effect on pumps/cgms
* Loop watches the `pumpManager.status.deliveryIsUncertain` and when true - it presents the view returned from `pumpManager.deliveryUncertaintyRecoveryViewController` in an alert; all pump managers just show the standard "pump settings" view; in our case - I added a message to the pump config root view when uncertain delivery is detected; and I'm not sure how to test/trigger it

the rest of the changes are just removing comments and minor tweaks/refactorings

### DIFF
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -19,6 +19,7 @@ protocol DeviceDataManager {
     var availableCGMManagers: [CGMManagerDescriptor] { get }
     var pumpManager: PumpManagerUI? { get }
     var pumpDisplayState: CurrentValueSubject<PumpDisplayState?, Never> { get }
+    var pumpManagerStatus: CurrentValueSubject<PumpManagerStatus?, Never> { get }
     var bolusTrigger: PassthroughSubject<Bool, Never> { get }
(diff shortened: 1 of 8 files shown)


Jon-b-m:
> * detectedSystemTimeOffset - we're supposed to report the offset to the pump manager, and loop does have an implementation - but none of the pump managers is using it, we are returning `0` (same as before `plugins`)
 
The time offset is used here: 

image

image

PR #1658, opened 2025-10-15
feat(autotune): add color indicators for basal rate increases and dec…

i3bdulr7man:
As discussed on Discord


Jon-b-m:
@i3bdulr7man  Thanks. Is the colour red, for a lower value, intuitive? And green, for a higher value? I'm afraid it might raise some eyebrows...

I think I would prefer to only display a changed value displayed as "⇢". The non-altered could be in .secondary colour, without the "⇢". With this change any additional colours (like red or green) should't be needed?

However I'm open to more suggestions, of course.

i3bdulr7man:
> @i3bdulr7man Thanks. Is the colour red, for a lower value, intuitive? And green, for a higher value? I'm afraid it might raise some eyebrows...
> 
> I think I would prefer to only display a changed value displayed as "⇢". The non-altered could be in .secondary colour, without the "⇢". With this change any additional colours (like red or green) perhaps wouldn't be needed?
> 
> However I'm open to more suggestions, of course.

@Jon-b-m Thanks for the feedback!
You make a good point — red and green could be interpreted differently depending on context. My initial idea was to make the changes more visually intuitive for the user (green = increase, red = decrease), since this pattern is familiar from data dashboards.

However, I agree that simplifying the view and highlighting only the changed values with the “⇢” symbol might provide a cleaner and more consistent look, or we could explore a subtle visual cue (like .secondary color for unchanged values) if you prefer.

Jon-b-m:
> However, I agree that simplifying the view and highlighting only the changed values with the “⇢” symbol might provide a cleaner and more consistent look, or we could explore a subtle visual cue (like .secondary color for unchanged values) if you prefer.

Yes, will you please update your PR to test? Please include a screenshot?

i3bdulr7man:
> @i3bdulr7man Thanks! Could you please remove these: image

Simulator Screenshot - iPhone 16 Pro - 2025-10-25 at 13 26 55

PR #1673, opened 2025-10-31
pump-related improvements

yurique:
* when a pump is not fully unboarded - show a "Re-connect pump!" message on the home screen

This seems to be a frequent issue with Medtronic pumps. After update to 8.0.0, the pump connection will be preserved, but it looks like the old Medtronic pump manager was not setting the "isOnboarded" flag in its state.

We didn't have a check for this condition on the home screen, so it would appear to the user like the pump is connected. But not everything would work - iAPS wouldn't try to refresh the state of the pump (reservoir, etc).

Now we will show the `Re-connect pump!` red message, tapping on it will open the pump selection screen.

<details>
  <summary>Testing</summary>

  * start adding a Minimed pump (in simulator),
  * finish the step where you enter the region and pump ID, tap `Connect` and then `Continue`,
  * but don't complete the `Pump clock` screen - swipe down the whole modal (not the `Cancel` button),
  * return to home screen,
  * the `Re-connect pump!` message should be displayed, 
  * tapping it should open the pump selection screen.
</details>

---

Additionally:
* remove the duplicate `KnownPlugins.pumpExpiration` function
* add Medtrum to known plugins - when reading reservoir

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/Header/PumpView.swift
+++ b/FreeAPS/Sources/Modules/Home/View/Header/PumpView.swift
@@ -47,150 +47,157 @@ struct PumpView: View {
         // let sim = state.pumpName.contains("Simulator") // Just For Testing
         HStack(spacing: 5) {
             // OmniPods and Medtrum nanos
-            if let date = expiresAtDate {
-                // Insulin amount (U)
-                if let insulin = reservoir {
-                    // 120 % due to being non rectangular. +10 because of bottom inserter
-                    let amountFraction = 1.0 - (Double(insulin + 10) * 1.2 / 200)
-                    if insulin == 0xDEAD_BEEF {
-                        if nano {
-                            medtrumInsulinAmount(portion: amountFraction)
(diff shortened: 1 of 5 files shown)


FreeAPS/Sources/Modules/Home/View/Header/PumpView.swift
```diff
@@ -47,150 +47,157 @@ struct PumpView: View {
...
-                                    if let timeZone = timeZone,
-                                       timeZone.secondsFromGMT() != TimeZone.current.secondsFromGMT()
-                                    {
-                                        ClockOffset(mdtPump: false)
+            if let pumpManager = state.deviceDataManager.pumpManager,
+               !pumpManager.isOnboarded
+            {
+                Text("Re-connect pump!").font(.statusFont).foregroundStyle(.red)
+                    .offset(y: -4)
+            } else {
```
yurique: only an additional if-else above, everything else is just indentation

PR #1674, opened 2025-11-03
main chart customizations and optimization

yurique:
# Main chart settings 

* moved home chart settings into a separate screen
* updated some settings labels (and translations)
* option to show/hide y-axis labels
* option to show/hide in-range background fill
* option to show/hide secondary chart backdrop
* option to show/hide predictions legend
* option to show/hide insulin activity grid lines
* option to show/hide insulin activity y-axis labels
* option to display BG peaks on the chart

<details>
  <summary>screenshots</summary>
  
  image
image
image

</details>

# Horizontal (glucose) grid lines and labels

The ticks are *not* `from ... to ... ` with a fixed number of steps anymore.
The following determines which horizontal lines (and labels) will be rendered.

* if `min glucose` is above `low threshold` - add a line only for min glucose [min glucose - smallest in the dataset)
* if `min glucose` is below `low threshold`, add `low threshold` as a line
* if `min glucose` is  more than 18 mg/dl below `low threshold` - add `min glucose` as well

* if `max glucose` is below high threshold - add a line only for `max glucose` [max glucose - greatest in the dataset)
* if `max glucose` is above high threshold, add high threshold as a line
* if `max glucose` is >18 mg/dl above `high threshold` - add lines at "nice" numbers above `high threshold` until `max glucose`

(this sounds complicated, but the goal is to not show grid lines too close to each other and show them at "nice" (round) numbers when possible)

<details>
  <summary>screenshots</summary>

    low threshold = `5` mmol/l, high threshold = `7` mmol/l
    min glucose = `3.3` mmol/l is below low threshold --> low threshold is added, min is more than 18mg/dl below low threshold - min glucose line is also added
    max glucose = `13.9` mmol/l  is above high threshold --> high threshold is added, lines between threshold and max are added with steps of `3` mmol/l at "nice" numbers (only `10 mmol/l` in this case), max glucose line is added

image

    low threshold = `4` mmol/l, high threshold = `10` mmol/l
    min glucose = `3.3` mmol/l  is below low threshold - low threshold is added, min glucose is not added (less than 18mg/dl from low threshold)
    max glucose = `13.9` mmol/l - the high threshold is added, lines between threshold and max are not added (max glucose too close to high threshold), max glucose line is added

image

</details>

# Prediction dots

Made prediction dots a little smaller because with colored BG points - predictions are hard to distinguish from BG history. Maybe worth making them even smaller?

# In-range background

Added an option to display the background rectangle around the in-range area of BG.

<details>
  <summary>screenshot</summary>
  image
</details>

Another option controls whether to render a background under the secondary chart:

<details>
  <summary>screenshot</summary>
image

image

image

</details>

#  BG peaks

* selecting peaks (min/max) in the glucose data
* displaying the peaks "inline" on the chart
* the algorithm tries to fit the labels such that they don't overlap with bolus or carbs
  * if can put directly above/below BG dot - ✅
  * otherwise, try moving the label to the right, left, up/down-right, up/down left, up/down
  * try moving further
  * if a valid spot not found within 3 previous and next BG dots - the peak is not displayed

<details>
  <summary>screenshots</summary>

image

image

image

image

image

image

image

image

image

image

---

image

IMG_5862

IMG_5863

IMG_5864

IMG_5865

</details>

# Main chart optimization

* previously, the "geometries" were computed multiple times per update (due to many different `.onChange(of: ...) { update(...)`)
* now we only update the "geometries" once
  * we "subscribe" to changes in the underlying data, but react to the changes a 100ms debounce
  * all calculations happen within a single block in `calculationQueue`
  * everything is calculated only once per data change / geometry change (debounced) - geometry is not used inside the view (except for vertical grid lines - to get the full height)
  * all state is assigned to `@State var`s within a single `DispatchQueue.main.async { }` - no multiple re-renderings for partially updated states
  * calculations are now optimized and take ~15-25 milliseconds on my phone; new calculations related to BG peaks - ~3 milliseconds

# TODO

* [x] peak placement needs a couple of tweks
* [x] peak placement needs some more testing (and probably tweaking) with carbs on the chart (bars and circles)
* [x] better in-range background color/opacity/etc - green doesn't look great (although it looks not too bad in light mode), but the regular gray didn't look nice either
* [x] add new localizable strings

### DIFF
--- a/FreeAPS/Sources/Modules/Home/View/Chart/CalculatedGeometries.swift
+++ b/FreeAPS/Sources/Modules/Home/View/Chart/CalculatedGeometries.swift
@@ -1,7 +1,5 @@
 import Foundation
-import SwiftUI
-
-import Foundation
+import SwiftDate
 import SwiftUI
 
 struct CalculatedGeometries {
@@ -1242,7 +1240,7 @@ private final class GeometriesBuilder {
 
     private static func calculateFirstHourDate() -> Date {
         let firstDate = Date().addingTimeInterval(-1.days.timeInterval)
-        return firstDate.dateTruncated(from: .minute)!
+        return DateInRegion(firstDate, region: .current).dateTruncated(from: .minute)!.date
     }
 
     private func calculateFirstHourPosition() -> CGFloat {


Jon-b-m:
Added some illustrations. Feel free to update:
image

PR #1708, opened 2025-11-29
Nightscout loop popup in daily graph

luborjurena:
Nightscout displays the loop popup on the daily graph only when the devicestatus document contains:
- openaps.suggested
- optionally openaps.enacted (but suggested is required for UI popup)

If your loop only populates enacted, Nightscout will not show the popup.
Additionally, devicestatus.created_at must be timestamp-aligned with the SGV (entries) timestamp to ensure proper placement on the graph.
Screenshot 2025-11-29 at 14 56 35


Jon-b-m:
In open loop nothing ever gets enacted, hence we shouldn’t upload a fake enacted, as that would display a false history of events in Nightscout.

Jon-b-m:
Looking OK now 👍, but perhaps remove comment on line 468 also, please.

luborjurena:
OK

PR #1742, opened 2025-12-21
ai rework

yurique:
### Summary:

* full refactoring of the AI and search-related code (cleaner architecture with protocols, no duplication, removed a ton of "dead" code, etc; probably 80% less code)
* full refactoring of the AI prompts (split into sections, extracted into *.txt assets, almost-typed output schema)
* the user can specify their preferred language (for descriptions, etc), their region and the nutrition standards authority
* more AI models (and fixed existing)
* refactored the UI
  * fixed some of the deprecations
  * simpler routing in the food search view (enum for visible screen, display as full-screen cover instead of navigation links, etc)
* refactored the UX flows 
  * camera button opens camera, long tap for "select from library"
  * can add multiple search results
  * edit detected portions, delete results/food items
  * etc etc
* nutrition values are no longer standard-serving-centric, rather per-100g-centric (standard serving sizes and multipliers are still displayed when available)
* simpler integration with the Add Carbs flow (removed the extra "take over" step)
* and more..

Sneak peek:
https://github.com/user-attachments/assets/2dfda769-f549-4aec-b080-9f9d3d68a406

### TODO:

* [x] saved Foods is now integrated.
* [x] finish cleanup (a few commented-out blocks are yet to be deleted; move/rename files/types)
* [x] tweak the prompt a bit more (sometimes the models go too far and instead of describing a "cheeseburger" they split it into ingredients as separate food items)
* [x] implement the "Save" functionality
* [ ] persist search results when the user returns from bolus calculator to the Add Carbs view
* [x] cancel button shows a prompt when there are search results, but swiping down just hides the view, no questions asked; not sure if this should also show a confirmation prompt
* [x] test menu/receipe photo analysis
* [x] test error state messages (invalid api key, bad response from ai, etc)
* [x] product images from database search results are available, but not included in the UI
* [x] settings screen cleanup
* [x] localizations
* [ ] ...

### Other future other

* ~~I'm thinking we should add an optional text field when analyzing a photo - to give AI some hints~~
Optional annotations are implemented.
* It might be possible to tell the models to include multiple foods when it's uncertain (when it's not sure if it's rise or spaghetti, etc) - and the user will then keep one and remove the rest
* AI models might produce better results if provided with two photos of the food, at different angles; although this will complicate the UI a bit


Jon-b-m:
@yurique, thanks. 

One small request: 
Could we use the lower summary (encircled in green) layout also at top (instead of red), please?

image

I think it would look more homogeneous this way.

Jon-b-m:
Yeah, I think I prefer the summary in carbs view and bolus view (non ai). It’s clear and clearly divided…

Jon-b-m:
Last suggestion is better 👍

yurique:
I pushed the latest version.

yurique:
Nice!
We can clean-up the `AddCarbs.RootView` even more (I tried not to touch this file too much up until now) - if we don't need the old saved-foods functionality (currently unused, but still there).

yurique:
> save as new preset button, which is missing now

yeah, you removed that one long time ago :)
https://github.com/yurique/iAPS/commit/15aa6e16db30a3dfc7c0e5f82efa9c9bbbd5647d

PR #1778, opened 2026-01-31
store CGM and pump state in binary plist files instead of UserDefaults.standard

yurique:
We've had reports from users about loosing their pump connection completely. 

https://discord.com/channels/1120154740857245808/1124373855813771286/1452757208466194453

@marionbarker mentioned this issue was fixed in Loop a couple of years ago by moving the device state from `UserDefaults.standard` into binary plist files. The corresponding PR is here: https://github.com/LoopKit/Loop/pull/1702/changes

This PR copies the `PersistedProperty` wrapper from Loop and uses it to store device data.
Same as Loop, it will try to read the old data from `UserDefaults.standard` if the new plist file does not exist.

Loop also deletes the old data from `UserDefaults.standard` after saving it into the plist, I implemented this as well, but it is currently commented out. 

I tested this on my phone - G7 sensor + Medtrum pump. The device settings got successfully transferred, the devices remained connected and working.

The plist files did appear in the iAPS directory:
image

### TODO
* [x] uncomment the cleanup of the legacy data (or keep the data there "just in case"?)

### DIFF
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -135,7 +135,7 @@ final class BaseDeviceDataManager: Injectable, DeviceDataManager {
             oldValue?.delegateQueue = nil
             setupCGM()
             rawCGMManager = cgmManager?.rawValue
-//            UserDefaults.standard.clearLegacyCGMManagerRawValue()
+            UserDefaults.standard.clearLegacyCGMManagerRawValue()
         }
     }
 
@@ -156,7 +156,7 @@ final class BaseDeviceDataManager: Injectable, DeviceDataManager {
 
             setupPump()
             rawPumpManager = pumpManager?.rawValue
-//            UserDefaults.standard.clearLegacyPumpManagerRawValue()
+            UserDefaults.standard.clearLegacyPumpManagerRawValue()
         }
     }
 


Jon-b-m:
@yurique ```Loop also deletes the old data from UserDefaults.standard after saving it into the plist, I implemented this as well, but it is currently commented out.```
I don’t see in what scenario this should matter? I think we can keep it as is.

yurique:
My thinking was: not deleting the legacy data allows to switch to a previous version of iAPS while keeping the device connected. For example, switch to `dev`, need to switch back to `main` for some reason - it will work.

But I gave it another thought, and realized that we still need to delete the data from the `UserDefaults` one way or another.
Because currently, if a device is removed - the file will get removed, but the data in `UserDefaults` will remain and the device will re-appear after restart, which will be at the very least very confusing.

I see two options:
* delete the data from `UserDefaults` immediately after writing it to the file (like Loop does)
* delete the data from `UserDefaults` when a pump or CGM data is set to `nil`

I'm okay with either of these.

* the first option is straightforward, because it's just one place for pump and one place for CGM
* the second one is not *too* bad either, but there are multiple places where we set `pumpManager` / `cgmManager` to `nil`, and in each of those places we'll need to also remove the "legacy" data (and we'll need to remember to keep them in-sync if/when something changes in the `DeviceDataManager`).

Jon-b-m:
@yurique ```I see two options:
delete the data from UserDefaults immediately after writing it to the file (like Loop does)
delete the data from UserDefaults when a pump or CGM data is set to nil```

What’s the reason not to do it like Loop?

PR #1836, opened 2026-04-02
Upload logs

scrappy:
This codes adds an *option* ( opt-in ) to Sharing that allows for auto-updating the previous days full logs to the server ... 

When logs are uploaded ( either manully after logging in, or automatically through the app ), a copy of the full logs are safed to disk, but a stripped down version is stored in the database ( removing all of the debug info )

When you either login, or go in via the Personal Statistics link, there is now a section for logs ... if you click there, you can see the cleaned up version of the log file for each day that is available, as well as an analyze link beside each log file that will run the log file through AI to give some insights.

ON the left menu, there is also a link for Analysis, which goes one step further and will run an analysis on log files going back in time, as long as there is no gaps.  

for instance, if there is a log for 20, 21, 22, 23, 25, 26, 27, 28, 30, 31, 01 ... it will use as far back as the 25th ... 

there is a lot of work to do on this ... the analysis will provide suggested tweaks to settings, with a full description of its reasoning for those suggestions ... I fed in the code, especially the algorithm, so that it has an idea of the logic involved ... 

one thing I want to build up is a graph that shows glucose over time, as well as points in time where tweaks were made to settings, so that you can visualize a 'before' and 'after' sort of thing

the idea of adding this code into the backend is that someone can choose to have their log auto-uploaded, so that there is as much data as possible available for analysis.   

it is designed to be conservative in its suggestions, but what I'm hoping for is that folks will provide input into the 'rules' ... for instance, on my first run ( 1 day worth ), it recommended changing autosens_max, but I thought there is someting about autoisf where you don't adjust that ... or am I miss remembering?

basically, hoping that as folks see errors in the analysis, we can discuss what they are and I can add to the logic to teach it to come up with better analysis.

this code is tested ( running on my phone right now ), and there is a 'Upload' button under Developer Options that you can use to seed yesterdays log file to give a head start.

### DIFF
--- a/FreeAPS/Sources/Logger/IssueReporter/SimpleLogReporter.swift
+++ b/FreeAPS/Sources/Logger/IssueReporter/SimpleLogReporter.swift
@@ -70,6 +70,12 @@ final class SimpleLogReporter: IssueReporter {
         getDocumentsDirectory().appendingPathComponent("logs/log_prev.txt").path
     }
 
+    /// Holds the log data for a day whose upload failed or could not be attempted.
+    /// Written when an upload is queued and deleted after a successful upload.
+    static var logFilePending: String {
+        getDocumentsDirectory().appendingPathComponent("logs/log_pending.txt").path
+    }
+
     static func getDocumentsDirectory() -> URL {
         let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
         let documentsDirectory = paths[0]
--- a/FreeAPS/Sources/Logger/IssueReporter/SimpleLogReporter.swift
+++ b/FreeAPS/Sources/Logger/IssueReporter/SimpleLogReporter.swift
@@ -39,6 +39,13 @@ final class SimpleLogReporter: IssueReporter {
                 try? fileManager.removeItem(atPath: SimpleLogReporter.logFilePrev)
                 try? fileManager.moveItem(atPath: SimpleLogReporter.logFile, toPath: SimpleLogReporter.logFilePrev)
                 createFile(at: startOfDay)
+                // Notify interested parties that log_prev.txt is ready for upload
+                let prevDate = Calendar.current.date(byAdding: .day, value: -1, to: startOfDay)!
+                NotificationCenter.default.post(
+                    name: .logDidRotate,
+                    object: nil,
+                    userInfo: ["logDate": prevDate]
+                )
             }
         }
 
@@ -86,4 +93,9 @@ private extension Data {
 
 private extension String {
     var file: String { components(separatedBy: "/").last ?? "" }
+}
+
+extension Notification.Name {
+    /// Posted after daily log rotation; `userInfo["logDate"]` is the Date the log covers (yesterday).
+    static let logDidRotate = Notification.Name("iAPS.logDidRotate")
 }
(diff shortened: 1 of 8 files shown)


Jon-b-m:
@scrappy there are two PRs. Is this the real?

scrappy:
Is there a reason why this patch cannot be applied?   It builds fine and I've been running it since I submit without any issues ... is there a bug with it?

Jon-b-m:
I haven't had time looking at this yet. 

But before looking at the code, there are perhaps some questions we need to address: 
1. Do we want to encourage our iAPS users to use the big AI-platforms for medical advice?
2. Can Claude AI adequately interpret the current iAPS code? From some of your examples it looks like it cannot? 
3. Can Claude AI or any other platform be trusted with changing your iAPS settings?

And before pushing this out to all iAPS users it needs to be tested. With the current tests I've seen from your posted examples it doesn't look like it's working, does it?

scrappy:
I think there is some confusion here ... the upload log function has nothing to do with AI or Claude ... it literally just uploads the log files to the server.   There is log parsing that happens, non AI, to clean out all of the garbage ( ie. debug lines ) to get it down to just the logs ( loop runs, bolus, carbs ) so that the result is easier to read as a human.

There are other *opt in* features that someone can make use of ... and the AI features definitely a case of 'click here to analyze' ... and nothing is fed back to iAPS ... *suggestions* are made, and very conservative ones at that.

But nothing goes to Claude automatically ... there isn't even a setting for "do this automatically" ... you have to login to your account ( if you created one ) and physically click a button askin gthe system to run an analysis.

Its no different then someone uploading all of their logs to Claude and asking the same thing, except that I'm trying to both simplify the process of saving those log files for submission, as well as encouraging input / adjustments to the data provided to AI in order to improvce those results in a method that benefits everyone.

Note that neither the Log Viewer nor the ISF Profile feature use AI ... the ISF Profile is meant to serve as a test bed where you can easily see the results for fixes for Autotune and Dynamic ISF ... but also provides a tool for those running older versions where such an improved autotune isn't available.   

But, this patch / code does *nothing* other then upload log_prev.txt to the server, and parse the log to create the log viewer ... everything else is "by request of the user" ... and technically this is too, as by default, even with this patch installed, nothing happens unless the patient goes into Sharing and turns it on ... its off by default.

Jon-b-m:
```I think there is some confusion here ... the upload log function has nothing to do with AI or Claude ```
- The logs are intended for AI-analysis, aren't they?

```Note that neither the Log Viewer nor the ISF Profile feature use AI ... the ISF Profile is meant to serve as a test bed where you can easily see the results for fixes for Autotune and Dynamic ISF```
- Where can the community review this code?

scrappy:
https://github.com/Artificial-Pancreas/iAPS/pull/1837

Posted a few days ago ... builds, haven't tested yet ... waiting on this patch then will switch my phone to use the autotune patch ...

scrappy:
if you are going to reject due to the ai features on the web site, then do so ... this code does not introduce any ai code to the app, and its all opt in ( you have to turn the upload feature on, you have to login, you have to add an AI token and you have to run the analysis ... nothing is automatic, and there are so many pop ups and warnings on the site to remind ppl that a) AI is involved and b) this is for informational purposes ) ... at every step, its a conscious decision by users to make use of, and the AI tool is just one aspect that makes use of the log files.

Since the AI Token itself costs money, I'm not expecting a lot of ppl to actually even make use of the AI features, but again, whether they do or not is their decision, just like everything about about the shared data aspect ... *shrug*

Jon-b-m:
Merged here https://github.com/Artificial-Pancreas/iAPS/commit/629f060bf6158d5209f24914b7206aeffc129ba8, thanks, along with some updated localizations.

PR #1837, opened 2026-04-02
Upgrade Autotune to fix issues with Dynamic ISF / Auto ISF issues

scrappy:
I didn't know the best way to share this, so ...

I passed this problem through Claude, to see what it would come up with ... it hasn't been tested, or even tried to build ... I wanted to throw the concept out there first ... if its going to be outright rejected because its "an AI solution", I don't want to spend a lot of time on it ...

I had it generate an AUTOTUNE_FIX.md file that gets into the details of what it figured out and what changes it made, so that its not *just code*, but the methodology behind it.

Even if not acceptable as it is, maybe it will give some ideas on changes that could be made ...

*If* the changes look good / reasonable, let me know and I will build / run the code to test it ...


Jon-b-m:
@scrappy  Do you have any tests/analysis to present?

scrappy:
I'm redoing this from a different angle, since I myself really don't like the idea of flat out replacing the current autotune, so am going to do this similar to how I vaguely recall the bolus calculator was redone .. in parallel vs straight replace

will re submit when I have something clean / ready

PR #1839, opened 2026-04-04
ai error logging, share logs in a zip file

yurique:
* replaced `print(...)` with `debug(.service, ...)` for ai errors
* "share logs" puts the logs into a zip file and shares the zip file

### DIFF
--- a/FreeAPS/Sources/Modules/Settings/SettingsStateModel.swift
+++ b/FreeAPS/Sources/Modules/Settings/SettingsStateModel.swift
@@ -66,6 +66,22 @@ extension Settings {
             subscribeSetting(\.animatedBackground, on: $animatedBackground) { animatedBackground = $0 }
         }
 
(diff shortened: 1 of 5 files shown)


FreeAPS/Sources/Modules/Settings/SettingsStateModel.swift
```diff
@@ -77,6 +93,47 @@ extension Settings {
...
+                try? FileManager.default.copyItem(at: url, to: dest)
+                zipURL = dest
+            }
+
+            try? fileManager.removeItem(at: stagingDir)
+
+            if let zipURL {
+                return [zipURL]
+            }
             return items
```
Jon-b-m: Could you fragment out this part into a zip function, please?

yurique: Done ✅

PR #1841, opened 2026-04-06
live activity on the watch

yurique:
This enables the iAPS live activity widged on the watch (smart stack).

The layout on the watch is tweaked a little bit, since we have less space, currently it looks like this:

* v2

incoming-9194E855-971B-4D08-A9F1-D7396509203A
incoming-9194E855-971B-4D08-A9F1-D7396509203A

<details>
  <summary>
  previous versions
  </summary>

* v1
incoming-9194E855-971B-4D08-A9F1-D7396509203A

</details>

This change should also make live activity on CarPlay display the full widget (with chart, etc), but I don't have a CarPlay-enabled car to test.

### DIFF
--- a/FreeAPS/Sources/Services/LiveActivity/LiveActivityBridge.swift
+++ b/FreeAPS/Sources/Services/LiveActivity/LiveActivityBridge.swift
@@ -264,12 +264,12 @@ final class LiveActivityBridge: Injectable, ObservableObject, SettingsObserver {
                         )
                         return ActivityContent(
                             state: state.withoutPredictions(),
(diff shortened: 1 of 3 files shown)


yurique:
I'm running it on my phone currently, to see if this layout works. And I'd say it needs tweaks.

* the BG indicator should be larger / easier to read
* IOB/COB - might be okay as is, but maybe worth making them a bit larger
* timestamp under BG - I don't like it

I'll try a couple more layouts. Probably tomorrow.

yurique:
Pushed an updated layout, updated the PR description.
All numbers are at the top row, larger font - seems to be pretty legible.

The timestamp is displayed only when the activity data is stale or last loop was more than 7 minutes ago.

PR #1857, opened 2026-04-24
Fix API paths

scrappy:
This is mostly a cleanup of the API endpoints on the server side, so that debugging is a bit easier ...

The old code ( server side is maintaining backward compatibility for old code right now ) had all uploads going into one end point 'upload.php', so there was no easy way to know what was being uploaded ... and this cleans up the URLs for download ( next project to be worked on )

Also, added insulinConcentration ( U100 vs U200 ) to pump_settings upload, since that was missing

Code builds fine, and is currently running on my phone ...

### DIFF
--- a/FreeAPS/Resources/javascript/prepare/determine-basal.js
+++ b/FreeAPS/Resources/javascript/prepare/determine-basal.js
@@ -285,5 +285,5 @@ function disableSMBs(dynamicVariables, now) {
 
 function isAISFenabled(profile) {
     const dynamicVariables = profile.dynamicVariables || { } ;
-    return (profile.iaps.autoisfEffective && !(dynamicVariables.aisfOverridden && !dynamicVariables.autoISFoverrides.autoisf)) || (dynamicVariables.aisfOverridden && dynamicVariables.autoISFoverrides.autoisf)
+    return (profile.iaps.autoisf && !(dynamicVariables.aisfOverridden && !dynamicVariables.autoISFoverrides.autoisf)) || (dynamicVariables.aisfOverridden && dynamicVariables.autoISFoverrides.autoisf)
 }
--- a/FreeAPS/Sources/Models/DatabaseModels.swift
+++ b/FreeAPS/Sources/Models/DatabaseModels.swift
@@ -5,6 +5,8 @@ struct DatabasePumpSettings: JSON {
     let settings: PumpSettings?
     let enteredBy: String
     let profile: String?
+    /// Insulin concentration factor: 1.0 = U100 (standard), 2.0 = U200, etc.
+    let insulinConcentration: Double?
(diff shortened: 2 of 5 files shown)


scrappy:
@Jon-b-m Sorry that its mixed together with my paths changes, but I'm not organized enough to deal with mulitple builds.

This last change above fixes History so that my Final is now properly showing AutoISF Max ( 3.5 ) and not 1 ...

scrappy:
---
  Bug: AutoISF ratio silently disabled — isAISFenabled() always returns false

  Commit e4cf8bdc3 ("Nighttime settings") introduced a regression that causes AutoISF to have no effect every loop cycle, regardless of settings. The AutoISF ratio is always 1 (inactive) even though autoisf.js runs correctly and computes the right ratio.

  Root cause

  The commit changed isAISFenabled() in prepare/determine-basal.js to check profile.iaps.autoisfEffective instead of profile.iaps.autoisf:

  // After the commit (broken):
  return (profile.iaps.autoisfEffective && ...) || ...

  autoisfEffective is a computed property on FreeAPSSettings in Swift:

  var autoisfEffective: Bool {
      autoisf && !isNighttime
  }

  Swift's auto-synthesized Encodable only serialises stored properties. Computed properties are never written to JSON. As a result, profile.iaps.autoisfEffective is always undefined in JavaScript — which is falsy — so isAISFenabled() always returns false.

  The effect: autoisf.js runs and correctly computes a ratio (e.g. 3.34), but determine-basal.js immediately discards it and falls through to Dynamic ISF or plain autosens instead.

  Fix

  Revert to the stored property:

  // Fixed:
  return (profile.iaps.autoisf && ...) || ...

  profile.iaps.autoisf is a stored Bool and is present in the JSON payload. The nighttime suppression logic already exists correctly in the Swift layer (OpenAPS.swift) — it doesn't need to be re-checked in JS.

  Impact

  Any build that includes commit e4cf8bdc3 has AutoISF effectively disabled at the JS layer, silently. The loop logs will show autoisf.js computing ratios normally, but determine-basal.js will log "Starting dynamic ISF layer." immediately after, indicating isAISFenabled() returned false.

Jon-b-m:
```Any build that includes commit https://github.com/Artificial-Pancreas/iAPS/commit/e4cf8bdc3206dd68a99deba686a33aa974a2c5d4 has AutoISF effectively disabled at the JS layer, silently. The loop logs will show autoisf.js computing ratios normally, but determine-basal.js will log "Starting dynamic ISF layer." immediately after, indicating isAISFenabled() returned false.```

 Fix pushed to dev branch. Please merge dev updates before continuing and remove the AI ”fix”.

Jon-b-m:
Fixed merge conflict

PR #1863, opened 2026-04-25
Autotune2.0

scrappy:
This does not replace autotune, but runs *with* it at this time.

Its meant to be a more robust ISF calculator then is currently in autotune, but since its brand new and needs more broad testing and feedback, it operators like the Basal Scheduler in Autotune, where you have to consciously decide to apply this to your profile ISF

Unless this is saved to your profile ISF, the values have no effect anywhere

It builds fine, and I currently have it running on my phone, so far with no problems .. the numbers come out what I think are normal, but feedback always good.


scrappy:
Removing this PR, as I'm working on a better one that is based on several papers that discuss how "less is better" in terms of time periods ... 

I was looking into whether it would make sense to increase the granularity to 30 minutes instead of 60, and found that the opposite is considered to be true ...

A full summary of the direction am going to go can be found:

   https://open-iaps.app/docs/granularity
   
With several sources referenced at the bottom ...

PR #1865, opened 2026-04-27
live activity - update no-chart banner for smart stack WIP

### LINKED ISSUE #1846: Issue with Apple Watch Live Activity
The new Live Activity on Apple Watch when display chart is disabled, shows the text oversized making impossible to read the BG values.
Reducing the text size to minimum doesn’t help.

Image

yurique:
fixes #1846

Unified the two views (for smart stack) - BG, COB, IOB and timestamp are the same.
The difference is the chart itself vs just eventual BG.

I also made the loop status and timestamp always visible on the chart view (previously it would only show when loop is "red"), and in the same location - bottom left.

The live activity updating in the smart stack is not as reliable as I thought it was, so having the timestamp seems to be necessary - otherwise one doesn't really know if the data is up to date.
Even though smart stack does display the "last updated ... minutes ago" above the widget - it appears for just a second, and seems to be rather random and unreliable (sometimes I see the up to date chart, but the overlay says "last updated 1 hour ago" etc).

---

I added `NSSupportsLiveActivitiesFrequentUpdates=true` to the plist hoping it would improve the reliability - it doesn't seem to have affected anything. Not sure if it's worth keeping.

---

* added eventual BG to the chart view in smart stack (same as in the no-chart view)
* removed the y-axis markers on the chart in smart stack (those are min/max of the BG values that are visible on the chart)

---

Image

Image

<br/>  

Image

Image

<br/>

Image

<details>
  <summary>previous screenshots</summary>

image

image

image

incoming-6C37F1FC-9F0B-48CA-A193-A6F9E652CD3D 2

</details>

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -74,7 +74,6 @@
 		198F67B82CFCBD5D00E538CB /* AutoISFStateModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 198F67B32CFCBD5D00E538CB /* AutoISFStateModel.swift */; };
 		198FA7BA2D380790006858FD /* ChartModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 198FA7B92D380790006858FD /* ChartModel.swift */; };
 		198FEED52D3F8C5900A67FAE /* LiveActivityChart.swift in Sources */ = {isa = PBXBuildFile; fileRef = 198FEED32D3E93CC00A67FAE /* LiveActivityChart.swift */; };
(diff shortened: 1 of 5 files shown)


yurique:
Made timestamp font larger.

Plus a small refactoring: extracted elements in different components into small `View` structs (from view `func`s).

Jon-b-m:
I think this look fine.

Jon-b-m:
@yurique please let me know when completed.

yurique:
@Jon-b-m I did a few more refactorings/cleanup (and updated the screenshots in the description to match the current code).

I think it's good to go now.

(I wouldn't hate it if more people tested this a bit :), but I guess that will be easier once it's merged)

PR #1866, opened 2026-04-30
Outdated version alert on Home screen

scrappy:
NOTE Build and Load works, but until next version is released, unable to actually test the pop up

Adds a safety-focused alert that fires when the running iAPS build is older than the current production release reported by the open-iaps version_check API (/api/v1/version_check → { main, dev }).

How it works
------------
- On app launch, HomeStateModel.subscribe() schedules a check 1.5 s after the view appears (so the alert doesn't interrupt startup).
- The same check is also wired into the existing 5-second DispatchTimer, so "Remind Me Later" dismissals re-surface automatically after 24 h without requiring an app restart.
- Version comparison is full semver (major.minor.patch), not just major.minor, so a patch bump on main (e.g. 8.2.0 → 8.2.2) correctly marks an 8.2.1 dev build as outdated.
- "Up to date" means device version >= main version, so users running the dev branch (e.g. 8.2.1 when main is 8.2.0) are not warned.

Dismissal behaviour
-------------------
- "Remind Me Later" — session-only: saves a timestamp to UserDefaults (iAPS.outdatedAlert.lastShown). The 24 h gate suppresses re-checks until the timestamp ages out; after that the timer triggers the alert again automatically.
- "Don't Remind Me Until Next Release" — persists a boolean keyed to the exact main version string (e.g. iAPS.outdatedAlert.main.8.2.0). When main increments to any newer version the key changes, so the alert automatically reappears for all users still running an older build regardless of prior dismissals.

The alert message explains the insulin-dosing safety rationale and directs users to the iAPS Discord (discord.com/invite/ptkk2Y264Z) for upgrade assistance.


Jon-b-m:
Do you have a screenshot?

Jon-b-m:
Please also add localizations

scrappy:
You are soooo needy ;)

I had to insert a false version record into version_check to trigger it, of course ... which is why it references v8.2.2 ... I've since removed that entry ...

Screenshot 2026-04-30 at 20 23 57

Jon-b-m:
@scrappy this can be merged after added localizations.

PR #1874, opened 2026-05-03
Upload fat and protein to Nightscout alongside carbs

scrappy:
When a meal entry includes fat or protein, those values were stored in CarbsEntry but explicitly dropped (set to nil) when building the NigtscoutTreatment for upload. Nightscout's /api/v1/treatments endpoint already accepts fat and protein fields on Carb Correction entries — the struct and JSON coding keys were already in place — so the fix is simply to pass the values through instead of discarding them.

FPU spread entries (future carb-equivalents generated from fat/protein content) carry fat: 0 / protein: 0 and are unaffected.


Jon-b-m:
@scrappy thanks. When fat are deleted or edited in iAPS we need to update also the NS fat/protein entries. Is this tested with iAPS deletion and edits?

yurique:
fat/protein are sent to Nightscout in the current `alpha` branch (as part of the treatments sync)

PR #1876, opened 2026-05-03
Search bar

### LINKED ISSUE #1852: Feature Request: Search Bar in Settings View
**Summary**
Add a search bar to the Settings view to allow users to quickly locate and navigate to a specific setting without manually browsing through sections.

**Motivation**
iAPS has a rich and extensive settings interface, which is a strength of the app — but it also means that finding a specific parameter (especially less frequently adjusted ones) often requires scrolling through multiple sections. A search bar would significantly reduce friction for both new users learning the system and experienced users who want to quickly verify or tweak a particular value.

**Proposed Behavior**
- A persistent or collapsible search field at the top of the Settings view
- Filters visible settings in real time as the user types
- Matches against setting names and ideally section headers
- Tapping a result navigates directly to that setting or highlights it in context

**Use Case**
A user wants to confirm or adjust a specific setting (e.g., a basal profile value, an algorithm parameter, or a notification threshold) without remembering which section it lives in. With a search bar, they type a keyword and land directly on the relevant control.

**Notes**
- Not sure if this is already in progress in a dev branch — happy to close this if so. If it hasn't been picked up yet, flagging it here for consideration.
- I'm not sure about the level of effort for some of these items, just trying to be helpful
- Used Claude to improve the feature description, hope it's clear and useful.

scrappy:
Builds and run / search works ... minimal testing though, as my live version is the autotune one, just flipped to this to make sure it actually works ...   

Closes #1852 

 Add search bar to Settings view

    Users with a specific setting in mind currently have to remember which
    section it lives in and scroll through up to 40+ options across 7 groups.
    A search bar significantly reduces that friction for both new users
    exploring the app and experienced users doing quick parameter checks.

    Implementation:
    - New SettingsCatalogue.swift: a static index of ~100 searchable entries
      across all settings areas (Devices, Services, Configuration, OpenAPS
      main/SMB/target/other, Autotune, Auto ISF, Dynamic ISF, Bolus Calculator,
      Fat & Protein, UI/UX). Each entry carries a display name, section label,
      and the Screen destination it navigates to.
    - SettingsRootView: adds .searchable() with displayMode .always so the
      field is immediately visible without pulling down the list. While the
      query is non-empty a results section appears above the regular settings
      showing each match with its setting name and a section subtitle. Tapping
      any result navigates directly to the containing settings screen via the
      existing .navigationLink(to:from:) mechanism. The normal settings layout
      is untouched when the search field is empty.

    Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -148,6 +148,7 @@
 		3811DE3225C9D49500A708ED /* HomeDataFlow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3811DE2A25C9D49500A708ED /* HomeDataFlow.swift */; };
 		3811DE3525C9D49500A708ED /* HomeRootView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3811DE2E25C9D49500A708ED /* HomeRootView.swift */; };
 		3811DE3F25C9D4A100A708ED /* SettingsStateModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3811DE3925C9D4A100A708ED /* SettingsStateModel.swift */; };
+		CC1A00012A000001000A708ED /* SettingsCatalogue.swift in Sources */ = {isa = PBXBuildFile; fileRef = CC1A00002A000001000A708ED /* SettingsCatalogue.swift */; };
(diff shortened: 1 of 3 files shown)


Jon-b-m:
@scrappy At a first glance (haven’t built yet) I think we need make following correction: 
1: use the localized strings when searching.

Jon-b-m:
@scrappy again I don't object to this idea, but currently it doesn't allow searches in native (user-preferred) language.

Issue #208, opened 2023-09-08
iAPS Profile Expiration Notification


AngryKiwiNZ:
HI Team, can we please have this feature built into iAPS to warn prior to certificate expiration (assuming it's not there already :) ).
https://loopkit.github.io/loopdocs/operation/features/notifications/#loop-app-expiration-notification

PR #1877, opened 2026-05-04
Add TestFlight build expiration alert

scrappy:
TestFlight builds expire 90 days after the build date. Users have no in-app warning today — the app simply stops launching once the deadline passes. This adds a Loop-style alert that fires when the home screen appears or the app returns to the foreground:

  ≤ 20 days remaining  → alert at most once every 2 days
  <  24 hours remaining → alert at most once per hour

BuildExpirationManager (new, Helpers/) computes the expiry as buildDate + 90 days using the existing Bundle.main.buildDate helper, gates alerts via a UserDefaults timestamp, and provides the alert title and message text. HomeRootView checks the manager on onAppear and on every .active scenePhase transition, then shows a native SwiftUI alert with an OK button and a "More Info" button that opens the iAPS releases page. Both new source files are registered in project.pbxproj.

Closes: #208


Jon-b-m:
@scrappy please localize!

scrappy:
done, but not sure where the conflict came from ... my dev is up to date in my fork, and I did a 'git merge dev' before I did any changes and it said it was all up to date ... yet obviously it wasn't?

PR #1880, opened 2026-05-04
Fix mmol/L display showing ~4884 after Nightscout backfill

scrappy:
Nightscout entries carry a raw Dexcom ADC value in the `unfiltered` JSON field (e.g. 88000 for a calibrated glucose of 88 mg/dL). The glucose display and delta calculation were both reading `unfiltered` and converting it with ×0.0555 (mg/dL → mmol/L), yielding absurd values like 4,884.0 mmol/L.

For real-time CGM readings DeviceDataManager sets `unfiltered` to the same mg/dL integer as `glucose`, so the bug was invisible in normal use and only surfaced after a backfill.

Fix both sites to use `glucose ?? sgv` (the calibrated, smoothed value) instead of `unfiltered`.


scrappy:
Pulling PR ... @yurique is going to submit one that fixes the NS integration itself, which should negate the need for this one ...

PR #1883, opened 2026-05-09
feat: some tweaks to Voice-Activated Meal Entry 

rsilvers129:
## Overview
Adds a dedicated voice-activated meal entry feature, improves Gemini AI response speed, and fixes macronutrient decimal display.
### 1. Voice-Activated Meal Entry 
- **SpeechRecognitionService** — On-device speech recognition using `SFSpeechRecognizer` with a robust multi-segment accumulator. Only fully-finalized text (`result.isFinal`) is locked into the buffer, preventing partial revision duplicates (like the "clan clam" bug). Automatically restarts the recognition task on pauses so users can speak indefinitely until tapping Done.
- **VoiceInputView** — Full-screen overlay with pulsing mic animation, live transcript display, and Done/Cancel/Retry controls.
- **Integration** — Mic button added to the Food Search toolbar (when AI is enabled) and Voice Input option added to the Home screen meal context menu. Transcribed text routes directly into the existing AI analysis pipeline.
- **Permissions** — Added `NSMicrophoneUsageDescription` and `NSSpeechRecognitionUsageDescription` to Info.plist.
### 2. Gemini AI Latency Reduction (`GeminiProtocol.swift`)
Added `responseMimeType: "application/json"` to the Gemini generation config. Forces the API into strict JSON output mode, preventing the model from generating conversational filler text. Significantly reduces response time for all AI food analysis modes (text, photo, barcode).
### 3. Macronutrient Decimal Formatting (`BolusRootView.swift`)
Formatted carbs, fat, and protein values in the Enact Bolus meal summary to a maximum of 1 decimal place. Previously, values like `80.9948 g` and `41.0015 g` were displayed with full arbitrary precision.

https://youtube.com/shorts/pjznx-UxUZs?si=7gzA_hL5PN6W2huC


yurique:
> Updated GeminiProtocol to explicitly request application/json responseMimeType

nice!

It looks like Claude also supports structured output: https://platform.claude.com/docs/en/build-with-claude/structured-outputs

Also it looks like we can also specify the output schema (instead of describing it in the prompt) (this would be a bigger effort and deserves a separate PR, but it might be worth looking into)

FreeAPS/Sources/Modules/AddCarbs/Models/FoodSearchStateModel.swift
```diff
@@ -304,6 +311,17 @@ import SwiftUI
...
+                // cancelled by another search task
+            } catch let AIFoodAnalysisError.networkError(error) {
+                if let urlError = error as? URLError, urlError.code == .cancelled {
+                    // cancelled by another search task
+                } else {
+                    try? await Task.sleep(for: .seconds(1))
+                    self.analysisStart = nil
+                    self.analysisEnd = nil
+                    self.analysisError = error.localizedDescription
+                }
```
yurique, line 324: is this necessary?

rsilvers129: No. I just fixed that.

yurique: I meant only these extra catch blocks :) 

As for the button itself - I don't have a strong opinion. It can be useful.

rsilvers129: > I meant only these extra catch blocks :)
> 
> As for the button itself - I don't have a strong opinion. It can be useful.

One nice thing about the button is it makes it obvious that it is one of the choices, along with photo, barcode, etc. It also requires one fewer screen tap to use.

I just pushed an update to fix the issue that existed with text accumulation.


BTW, I now think voice description of a meal gives more accurate results than a photo.

rsilvers129:
I have been testing this for two weeks now and it works well. No issues. Also I like the dedicated voice button. Not just that, but I have found that voice is easier and faster, more accurate, and less AI expensive than taking a photo.

Jon-b-m:
@rsilvers129 Thanks for the PR. I will test it during this coming week. Can you please first remove your developer ID (you appear to have signed the project before pushing the PR) from this PR? And also please look into resolving the current merge conflicts.

Jon-b-m:
I now fixed the merge conflicts!

Jon-b-m:
@rsilvers129 While testing your PR I get these warnings: 
```
AVAudioSession_iOS.mm:978 This method can lead to UI unresponsiveness if called on the main thread. Consider using the asynchronous activate/deactivate API instead for calls from the main thread.
SessionCore.mm:631 This method can lead to UI unresponsiveness if called on the main thread while the audio session is active.
```

Jon-b-m:
Resolved the warnings and removed your bundle id everywhere.

Issue #1864, opened 2026-04-26
GitHub build failing at Fastlane gym step


vaksi24:
Hi, 

<img width="2343" height="1227" alt="Image" src="https://github.com/user-attachments/assets/fa135e53-ad0e-40d1-a8c0-fb4e31fd94b2" />

Build iAPS GitHub Action fails repeatedly at Fastlane gym step.

Fork synced with upstream main (8.2.0)
Repeated builds #5-#9 fail.

Certificates checks pass.

Error:
- Error building the application
- Fastlane finished with errors
- fastlane 2.233.0 is available, runner uses 2.228.0

Looks like GitHub build pipeline issue.

Screenshot attached.

Thanks.

yurique:
> ::error file=/Users/runner/work/iAPS/iAPS/FreeAPS.xcodeproj::Provisioning profile "match AppStore ru.artpancreas.***.FreeAPS.watchkitapp" doesn't support the HealthKit capability.  (in target 'FreeAPSWatch' from project 'FreeAPS')

You need to update the `FreeAPSWatch` identifier on the Apple Developer portal:
https://developer.apple.com/account/resources/identifiers/list

Find your `FreeAPSWatch` in the list and enable `HealthKit` for it

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/79bb8a07-6109-4335-a22f-f04fe9473fff" />

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/6eed7f4a-091e-458c-b9f8-be2270e0e36d" />

vaksi24:
Solved — thank you very much!

The issue was exactly the FreeAPSWatch identifier missing HealthKit capability in Apple Developer.

Enabled HealthKit for the watch identifier, rebuilt, and build #10 uploaded successfully to TestFlight. Updated to 8.2.0 successfully.

Really appreciate the quick help!

PR #1884, opened 2026-05-09
Fastlane: enable HealthKit capability on FreeAPSWatch (watchkitapp) App ID

shward:
## Summary

- Add `HEALTHKIT` to the `configure_bundle_id` call for `FreeAPSWatch` (`watchkitapp`) in `fastlane/Fastfile` so `fastlane identifiers` enables HealthKit on the App ID automatically.
- Fixes the Xcode 26 build failure where `gym` errors out with: `Provisioning profile "match AppStore <bundle>.watchkitapp" doesn't support the HealthKit capability` / `doesn't include the com.apple.developer.healthkit` entitlement.
- Without this, every forker has to manually enable HealthKit on the watch App ID in the Apple Developer portal (see #1864) before TestFlight builds will succeed.

## Why this is needed

The watch extension target (`watchkitapp.watchkitextension`) already has HEALTHKIT in its `configure_bundle_id` call, but starting with Xcode 26 the parent watch container target (`watchkitapp`) is also required to have the HealthKit entitlement — its provisioning profile is checked by `xcodebuild -exportArchive`, and the build fails if the App ID was created without the capability. Several users hit this on 8.2.0 (#1864); the documented manual workaround is to tick the HealthKit box on the `FreeAPSWatch` identifier in the Apple Developer portal. Codifying it in the `identifiers` lane makes that automatic on first build.

## Test plan

- [ ] Run the `2. Add Identifiers` workflow on a fork whose `FreeAPSWatch` App ID does not yet have HealthKit — verify the capability is enabled afterwards.
- [ ] Run the `3. Create Certificates` workflow (with `ENABLE_NUKE_CERTS=true` if needed) so match regenerates the watchkitapp provisioning profile.
- [ ] Run the `4. Build iAPS` workflow — verify it completes through `gym` and uploads to TestFlight without the HealthKit capability error.
- [ ] Re-running `2. Add Identifiers` on a fork that already has HealthKit enabled is a no-op (idempotent).

Refs: #1864

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -189,7 +189,8 @@ platform :ios do
     ])
     
     configure_bundle_id("FreeAPSWatch", "#{APP_IDENTIFIER}.watchkitapp", [
-      Spaceship::ConnectAPI::BundleIdCapability::Type::APP_GROUPS
+      Spaceship::ConnectAPI::BundleIdCapability::Type::APP_GROUPS,
+      Spaceship::ConnectAPI::BundleIdCapability::Type::HEALTHKIT
     ])
 
     configure_bundle_id("LiveActivityExtension", "#{APP_IDENTIFIER}.LiveActivity", [


Jon-b-m:
@shward thanks! Please let me know when tested.

PR #1885, opened 2026-05-09
enforce max basal rate when enacting a manual temp basal

yurique:
Fixing the issue reported in [Discord](https://discord.com/channels/1120154740857245808/1124371739544125522/1502752963477770366)

* before enacting a manual TBR (in Open Loop) - validate it, do not allow values greater than the Max Basal setting
* if it is greater - show an error message

Additionally:
* added a clamp to the `APSManager.swift#enactTempBasal` - even if a TBR greater than max is requested (which would be a bug somewhere else) - clamp it to Max Basal
* removed the filtering of pump manager events in `DeviceDataManager` - previously, it would filter out TBRs greater than Max Basal (looks like a workaround for an issue from years ago) - this was masking the issue and the user would not see the large manual TBR in the pump history (nor on the chart) after accidentally enacting it.

image

### DIFF
--- a/FreeAPS/Sources/APS/APSManager.swift
+++ b/FreeAPS/Sources/APS/APSManager.swift
@@ -516,6 +516,9 @@ final class BaseAPSManager: APSManager, Injectable {
             return
         }
 
+        let maxBasal = Double(settingsManager.pumpSettings.maxBasal)
+        let rate = duration > 0 ? min(rate, maxBasal) : rate
+
         debug(.apsManager, "Enact temp basal \(rate) - \(duration)")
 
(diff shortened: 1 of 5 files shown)


yurique:
Testing:

* tried enacting a large TBR - got an error message
* tried enacting a TBR=Max Basal - successful
* tried enacting a TBR less than Max Basal - successful 
* tried canceling a TBR - successful

Jon-b-m:
Thanks @yurique! Did you do the testing on Medtrum or Omnipod?

yurique:
I tested on the simulator - but the changes are not pump-specific, just iAPS code.

PR #1888, opened 2026-05-12
live activity - settings for smart stack (watch)

yurique:
It's hard to make the layout work correctly on different watch sizes. Some users prefer to see the BG delta, with and without chart, etc - and on smaller watches there's just not enough room to fit everything.

This PR introduces a couple of settings for the live activity in the Smart Stack (watch).

* `Watch: Show Delta` - whether to show BG delta on the watch (both no-chart and with-chart views)
  * when the chart is enabled - BG delta is displayed above the eventual BG; with chart disabled - BG delta is displayed in the bottom row, before the eventual BG
* `Watch: Show Chart` - when the chart is enabled, the user can disable it only on the watch (but keep on the phone lock screen)
* `Watch: Show Predictions` - when predictions are enabled, the user can disable them on the watch (but keep on the phone lock screen)
* `Watch: Show Eventual` - whether to show eventual BG on the watch with chart enabled; if eventual is disabled - the timestamp is moved to the right side of the widget - to not cover the chart (which was one of the reported problems on a smaller watch)

The screenshots below are from the simulator, and it might seem that all views look good, but on smaller watches there's much less space - vertically and horizontally - and some combinations won't work well, but the user can now adjust the settings to their needs.

image

image

image

image

Settings (GIF, click to play):

CleanShot 2026-05-12 at 15 18 57


yurique:
Added a setting to hide the eventual BG from the watch+chart view (updated the PR description)

Jon-b-m:
Thanks, @yurique, will test soon.

PR #1890, opened 2026-05-19
Features/advanced statistics

Rig22:
## Summary

Inspired by Trio Statistics

Redesigns the Statistics module with five tabs and dedicated chart
types per tab. The existing tab entry point (`state.showModal(for:
.statistics)`) is unchanged — the new module is a drop-in replacement
for the old `Stat.RootView`.

## Tabs

- **Overview** — 2×2 mini-donut landing page (Glucose / Insulin /
  Looping / Meals). Each donut is tappable for details, each card
  navigates to the full tab.
- **Glucose** — chart-type switch between
  - **Sector & Metrics** — TIR donut with Low / In-Range / Tight-Range
    / High tiles + eA1c, SD, CV, days.
  - **Percentile (AGP)** — smooth continuous 10-90 / 25-75 bands plus
    median line.
  - **Distribution** — stacked bar chart with clinical ranges
    (<54 / 54-70 / 70-180 / 180-250 / >250).
  All chart types work for Today / D / W / M / 3M.
- **Insulin** — TDD trend chart + bolus distribution chart, plus
  hero/tile summary with yesterday / 2-days / 3-days TDDs and
  comparison deltas.
- **Looping** — loop performance bar chart with success-rate badge,
  median duration, and most-frequent-error popover.
- **Meals** — carbs/protein/fat bar chart + macro-distribution donut
  with kcal tiles and reference legend.

## Design

- 4-card layout per tab (chart card + summary card)
- Tile pattern with icon + value + sublabel, rounded-corner outlines
- Tap-to-select popovers on bar charts (date + value)
- ContentUnavailableView empty states everywhere data can be missing
- Uses existing iAPS asset colors only (`loopGreen`, `loopYellow`,
  `loopRed`, `insulin`, `lightBlue`, `darkerBlue`, `purple`) — no new
  design tokens.

## Localization

All user-visible strings wrapped with `NSLocalizedString`. New keys
need to be added to the Crowdin source `Localizable.strings` so the
20+ existing translations can pick them up.


Jon-b-m:
Thanks @Rig22 . I plan to test your PR soon. On a quick glance it looks like you're continuously computing all of the glucose statistics..This seems unnecessarily taxing. Instead preferably you only need to compute this once at every new reading.

Rig22:
@Jon-b-m 

Good Morning Jon. :)

Thanks for taking a look! Quick clarification on the lifecycle — the
stats aren't computed continuously. StateModel.subscribe() runs
exactly once when the Stat view is opened, and inside it the four
setupXxx() methods each do a single CoreData fetch + aggregation:

setupInsulinStats() — 90 d of InsulinDistribution
setupInsulinSummary() — 10 d of TDD
setupMealStats() — 90 d of Carbohydrates
setupAGP() — 90 d of Readings, with AGP slots + distribution
precomputed for all five intervals (Today / D / W / M / 3M)
Switching tabs or changing the Duration picker afterwards only
re-filters arrays that already live in memory — no fetches, no
re-aggregation. So the cost is one upfront pass on entering the
view, then ~free for the rest of the session.

Two improvements I'm happy to add if you'd still like to reduce
that upfront cost:

Move the four setup calls onto a background queue and publish
on DispatchQueue.main when done — the view appears instantly
with empty cards that fill in within ~100–300 ms. Same fetch
work, just off the main thread.
Lazy per-interval aggregation — only compute the AGP /
distribution slots for the currently selected interval, and
cache on demand when the user switches. Saves ~80 % of the AGP
aggregation work for users who never open the longer windows.
Or, if you'd prefer the compute-on-new-reading pattern, that would
mean a long-lived observer running every ~5 min Loop cycle even
when the Stat view isn't open. Happy to go that route, just wanted
to flag the trade-off (battery + persistent state vs. one-shot
on-demand work).

Let me know which direction you'd prefer and I'll push the change.

Greetings Richard

Jon-b-m:
Right, I see now after some debugging that the complex stat computations are only computed when tapping the buttons, even though called directly from SwiftUI (which is why I was suspecting extra processing)…

Rig22:
The new Nutrients Push completely clears the meal data display within the Advanced Statistics > Meals view. This needs to be fixed, but I don't have time for it right now.

Jon-b-m:
> The new Nutrients Push completely clears the meal data display within the Advanced Statistics > Meals view. This needs to be fixed, but I don't have time for it right now.

This is temporary. The Carbohydrate Entity has been removed. When you have more of Meals data you should be able to display data again.

PR #1905, opened 2026-06-10
remove OmniBLE, OmniKit submodules, replaced with OmnipodKit

yurique:
This PR removes the old `omni-*` git submodules and the corresponding references from the code, workspace and from `FreeAPSWorkspace` scheme.

The UI will only show the new `All Omnipod Types` pump in the list.

If you had an existing omni pump connected before updating to this version (using one of the old plugins) - you will be seamlessly transitioned to the new pump manager on app start.
This is implemented the same way Loop did it: https://github.com/LoopKit/Loop/pull/2426/changes
* [ ] this should be tested

---

#### Notes:

* before building this branch, you need to perform the `Product` / `Clean Build Folder ...` action.

Otherwise, even though the old submodules are no longer present, nor visible in the workspace or in the code, the build directory can still contain the previously built artifacts for those plugins (this was the case for me, at least; causing some unnecessary confusion); and thus, the supposed-to-be-removed plugins will be picked up by the plugin manager and they will show up in the pump selection menu (and they will get activated to manage the pump, if there was a configured `omni-*` pump).

* I updated the `.swiftformat` file to prevent swift formatter from auto formatting all code in the `OmnipodKit` directory - if you previously built iAPS using this branch - you need to clean it up:
```sh
$ cd OmnipodKit/
$ git reset --hard HEAD
```

Unfortunately, there is another issue with files inside the `OmnipodKit` submodule getting modified and producing local uncommitted changes:
* xcode modifies the localization files in `OmnipodKit` on build (during the "Syncing Localizations" step):
```sh
$ git status
...
modified:   OmnipodKit (modified content)
$ cd OmnipodKit
$ git status
...
modified:   Localization/Localizable.xcstrings
```
I don't currently know how to fix this issue.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,12 +13,6 @@
 [submodule "MinimedKit"]
 	path = MinimedKit
 	url = https://github.com/Artificial-Pancreas/MinimedKit.git
-[submodule "OmniBLE"]
-	path = OmniBLE
-	url = https://github.com/Artificial-Pancreas/OmniBLE.git
-[submodule "OmniKit"]
-	path = OmniKit
-	url = https://github.com/Artificial-Pancreas/OmniKit.git
 [submodule "dexcom-share-client-swift"]
 	path = dexcom-share-client-swift
 	url = https://github.com/Artificial-Pancreas/dexcom-share-client-swift.git
--- a/.swiftformat
+++ b/.swiftformat
@@ -11,3 +11,4 @@
 --exclude RileyLinkKit/
 --exclude NightscoutRemoteCGM/
 --exclude MedtrumKit/
+--exclude OmnipodKit/
--- a/FreeAPS.xcworkspace/contents.xcworkspacedata
+++ b/FreeAPS.xcworkspace/contents.xcworkspacedata
@@ -8,12 +8,6 @@
       location = "group:RileyLinkKit/RileyLinkKit.xcodeproj">
    </FileRef>
    <FileRef
-      location = "group:OmniKit/OmniKit.xcodeproj">
-   </FileRef>
-   <FileRef
-      location = "group:OmniBLE/OmniBLE.xcodeproj">
-   </FileRef>
-   <FileRef
       location = "group:MinimedKit/MinimedKit.xcodeproj">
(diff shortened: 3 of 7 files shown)


yurique:
Sidenote:
`OmnipodKit` submodule should probably be switched to the `dev` branch before merging into dev.

Jon-b-m:
@yurique thanks! I can test this this later this week. The submodule branch will be updated whenever necessary code is merged into dev.

yurique:
@Jon-b-m right, I was thinking about a different thing and mixed things together.

What I meant to say is this:
* the old modules will no longer be maintained (most likely)
* Loop has switched to the new `OmnipodKit` plugin for DASH and Eros already (without O5 for now)
* so I'm thinking we should do the same - merge this into `dev`, with `OmnipodKit` pointing to the `dev` branch
* as soon as O5 support is finalized - it will land in the `dev` branch as well
* whether to keep the `omnipod5` branch - I'm not sure; people should not be trying to use O5 as of now, unless they are participating in the "final testing campaign" and collaborating with the dev team

Jon-b-m:
> @Jon-b-m right, I was thinking about a different thing and mixed things together.
> 
> 
> 
> What I meant to say is this:
> 
> * the old modules will no longer be maintained (most likely)
> 
> * Loop has switched to the new `OmnipodKit` plugin for DASH and Eros already (without O5 for now)
> 
> * so I'm thinking we should do the same - merge this into `dev`, with `OmnipodKit` pointing to the `dev` branch
> 
> * as soon as O5 support is finalized - it will land in the `dev` branch as well
> 
> * whether to keep the `omnipod5` branch - I'm not sure; people should not be trying to use O5 as of now, unless they are participating in the "final testing campaign" and collaborating with the dev team

Right, in that case it would be better to (also) PR dev directly with this new OnnipodKit dependency. And later when O5 is ready, we can merge also this branch into dev...

yurique:
PR into dev: https://github.com/Artificial-Pancreas/iAPS/pull/1908

Jon-b-m:
Tested

PR #1908, opened 2026-06-10
[dev] remove OmniBLE, OmniKit submodules, replaced with OmnipodKit

yurique:
This PR removes the old `omni-*` git submodules and the corresponding references from the code, workspace and from `FreeAPSWorkspace` scheme.

The UI will only show the new `All Omnipod Types` pump in the list.

If you had an existing omni pump connected before updating to this version (using one of the old plugins) - you will be seamlessly transitioned to the new pump manager on app start.
This is implemented the same way Loop did it: https://github.com/LoopKit/Loop/pull/2426/changes
* [ ] this should be tested

---

#### Notes:

* before building this branch, you need to perform the `Product` / `Clean Build Folder ...` action.

Otherwise, even though the old submodules are no longer present, nor visible in the workspace or in the code, the build directory can still contain the previously built artifacts for those plugins (this was the case for me, at least; causing some unnecessary confusion); and thus, the supposed-to-be-removed plugins will be picked up by the plugin manager and they will show up in the pump selection menu (and they will get activated to manage the pump, if there was a configured `omni-*` pump).

* I updated the `.swiftformat` file to prevent swift formatter from auto formatting all code in the `OmnipodKit` directory - if you previously built iAPS using this branch - you need to clean it up:
```sh
$ cd OmnipodKit/
$ git reset --hard HEAD
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,12 +13,6 @@
 [submodule "MinimedKit"]
 	path = MinimedKit
 	url = https://github.com/Artificial-Pancreas/MinimedKit.git
-[submodule "OmniBLE"]
-	path = OmniBLE
-	url = https://github.com/Artificial-Pancreas/OmniBLE.git
-[submodule "OmniKit"]
-	path = OmniKit
-	url = https://github.com/Artificial-Pancreas/OmniKit.git
 [submodule "dexcom-share-client-swift"]
 	path = dexcom-share-client-swift
 	url = https://github.com/Artificial-Pancreas/dexcom-share-client-swift.git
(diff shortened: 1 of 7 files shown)


yurique:
@Jon-b-m the `OmnipodKit` submodule is pinned to the latest commit from upstream dev branch (but the submodule points to https://github.com/Artificial-Pancreas/OmnipodKit - so git won't be able to find the commit).
We'll need to pull the dev branch from upstream and push to `Artificial-Pancreas/OmnipodKit`. The upstream dev branch doesn't need any tweaks - swift format will not touch it and the build no longer messes with localization files.

Jon-b-m:
Closed in favour of this: https://github.com/Artificial-Pancreas/iAPS/commit/95662251cb33ded014aef8ad4c19965e8600c73d

PR #1949, opened 2026-07-01
Onboarding & Restore — unified first-launch setup flow

scrappy:
## Onboarding & Restore — unified setup flow

This continues the cloud-restore work (the Phase A–C restore + account login from the already-merged #1941) into a complete first-launch onboarding experience. Opening fresh against `dev` since the branch lineage got tangled with that merge — happy to reorganize/rebase however suits.

### What it does

One onboarding step machine for **both** new and existing users — they differ only in whether each screen starts from restored data or from defaults:

```
Welcome
  → [Existing: restore by recovery token or account login]
  → Sharing (enable Online Backup under the new device id)
  → [Existing: CoreData preset restore status]
  → Device Setup (pump + CGM)
  → Software Setup summary
  → Setup Complete (Closed Loop)
  → Recovery Token & Account
```

Highlights:

- **Existing-user restore** is silent and all-or-nothing on the dosing-critical core (OpenAPS preferences, app settings, pump config incl. insulin concentration, and the basal/ISF/CR/target schedules), with best-effort extras (Contact Trick, AI settings, presets). Fixes a dealloc-cancel bug where fire-and-forget CoreData preset fetches were cancelled before they saved.
- **CoreData preset restore-status** screen reports meal / override / temp-target counts from an explicit, awaited restore (always shown so a missing backup is visible).
- **Software Setup summary** reuses iAPS's real editors via the Router — no parallel editors, so new upstream settings appear automatically. New users get a **Setup → Review** checklist (rows flip to "Review" once opened) so a knowledgeable helper can skip editors that don't apply.
- **Device Setup** step (pump + CGM) reflects real pairing state from `DeviceDataManager`.
- **Recovery Token & Account** final step shows the device's recovery token to record, with an explicit note that **an account is optional** — the token alone backs up and restores. Optional, opt-in link-to-existing-account (login, honoring 2FA) or create-account. Default action is "Finish without an account", so nothing links unless the user chooses to.

### Localization

Following @Jon-b-m's review — thank you, this was a real bug and a subtle one.

The keys were all present in the `en.lproj` source catalog, but four call sites never reached them. `Text`, `Label` and `navigationTitle` each have both a `LocalizedStringKey` and a plain-`String` initialiser, and a bare `cond ? "A" : "B"` of two string literals types as `String` — so it binds the **verbatim** initialiser and ships English-only, silently, with the translation sitting unused in the catalog:

- `DeviceSetupView.deviceRow` — "Review" / "Setup"
- `RecoveryTokenAccountView` — "Copy" / "Copied", and "Done" / "Finish without an account"
- `RestoreSummaryView` — "Set up your settings" / "Review your settings" navigation title

All four now hoist to an annotated `LocalizedStringKey` local before use. (A ternary passed to a parameter that is *already declared* `LocalizedStringKey` is fine — the contextual type wins; it's only call sites where both overloads are visible that break.)

Also added 9 keys that were used but had never been appended to the source catalog, so Crowdin had never seen them: the Sharing weight/height unit abbreviations (`kg`/`lb`/`cm`/`ft`/`in`/`ft/in`, with a translator comment for context) and the Developer-section replay-onboarding rows.

### Backup completeness

Auditing what the backup covers against what the device actually persists turned up three gaps — all of them things the CoreData model and settings surface grew past the backup payload, rather than anything that was ever deliberately excluded:

- **AI settings.** The food-search / photo-analysis settings live in `UserDefaults` (`UserDefaults+AI.swift`), not in `freeaps_settings.json`, so they were outside the backup entirely. The twelve preference keys now round-trip.

  The four provider API keys (Claude, OpenAI, Gemini, USDA) are **deliberately excluded** and should stay that way. They're the user's own billable third-party credentials, and putting them on a backup server makes that server a credential store whose compromise costs users money directly. Re-entering a key after a restore is the cheaper trade. `MigratedAISettings` documents this at the definition so it isn't quietly undone later.

- **Meal presets were being truncated.** `Presets` has grown to 16 attributes plus a micronutrients relationship, but `MigratedMeals` still carried four (`dish`/`carbs`/`fat`/`protein`). A restore wasn't merely failing to bring the rest back — it was handing back a *stripped* preset, dropping fibre, sugars, glycemic index, portion/serving size, tags, image and micronutrients. Now round-trips in full; micronutrient definitions are looked up by name on the restoring device and created if unknown.

- **Override presets lost their Auto ISF block.** `Auto_ISF` is a separate entity keyed by the preset's `id`, so it never travelled with the preset — a restored preset came back with its per-preset Auto ISF configuration gone, which is dosing-relevant. Now carried alongside each preset and written as its own row.

Everything added to the wire format is `Optional`. Swift's synthesized `Decodable` throws on a missing key rather than falling back to a property's default, so `Optional` is what keeps backups written by older builds decoding — and a nil field deliberately leaves that setting alone rather than overwriting it with a zero.

### Verification

- ✅ **GitHub Actions build: green** — `Build iAPS (restore-account-tokens)`, run `30177436603`.
- ✅ **Installed from TestFlight and running on device.**
- ✅ **Xcode simulator tested** — existing-user restore (settings + meal/override presets verified restored), new-user setup walk-through, device setup, and account creation (new account + device link confirmed landed in the backend database).
- ✅ **Server endpoints smoke-tested live** — AI settings and full meal presets verified round-tripping upload → download, and backups written by older builds still serve correctly through the fallback path.
- Synced with current `dev` (through `77cb19d8d`); merges clean.

One thing stated plainly: the client half of the three backup additions above is compiled and the app runs, but the restore of those specific new fields hasn't yet been exercised end-to-end on a device against a backup containing them. The server half is verified. Happy to hold that part back into a follow-up PR if you'd rather this one stay scoped to the onboarding flow.

The account link/create endpoints and the backup/restore sections are backed by companion changes on the open-iaps.app server, already deployed; the recovery-token backup/restore path itself works without an account.


Jon-b-m:
So we can delete the onboarding branch?

scrappy:
Yes ... thx

Jon-b-m:
@scrappy Do you have a list of which settings are restored and which still need additional work/extensions?

scrappy:
All settings that are being backed up are being restored, and there is at least one that wasn't being backed up that has been added ...

Jon-b-m:
> All settings that are being backed up are being restored, and there is at least one that wasn't being backed up that has been added ...

Good

Jon-b-m:
@scrappy please make some small updates to support localisation. Currently you are passing unlocalised variables (strings)into the SwiftUI deviceRow function. These strings are currently not localised.

scrappy:
@Jon-b-m Localisation is fixed, and the AI settings question turned out to be the thread that unravelled two more gaps. All pushed — build 30177436603 is green and I've run it on device.

Localisation. Good catch, and a subtler bug than it looked. The keys were all in en.lproj — the call sites just never reached them. Text, Label and navigationTitle each have both a LocalizedStringKey and a plain-String initialiser, and a bare cond ? "A" : "B" of two literals types as String, so it binds the verbatim initialiser and ships English-only with the translation sitting unused. Four sites: deviceRow's Review/Setup (the one you spotted), Copy/Copied and Done/Finish-without-an-account in RecoveryTokenAccountView, and the Set-up/Review navigation title in RestoreSummaryView. All now hoist to an annotated LocalizedStringKey local first. Worth knowing generally, since it fails silently and a catalogue check won't catch it.

I also found 9 keys that were used but had never been appended to the source catalogue, so Crowdin had never seen them — the Sharing weight/height unit abbreviations and the Developer replay-onboarding rows. Those are in now, with a translator comment on the units since in and ft are meaningless without context.

AI settings. You were right, they're in UserDefaults (UserDefaults+AI.swift), completely outside freeaps_settings.json and so outside the backup. The twelve preference keys now round-trip.

I deliberately left out the four provider API keys (Claude, OpenAI, Gemini, USDA), and I'd argue they should stay out. They're the user's own billable third-party credentials — backing them up makes the backup server a credential store, and a compromise there costs users money directly rather than just leaking a max-basal figure. Re-pasting a key after a restore is the cheaper trade. It's commented at the definition so it doesn't get quietly undone later. Happy to be overruled if you see it differently.

Two more gaps, and one of these matters more than the AI settings. Both are cases where the CoreData model grew past the backup payload:

- Meal presets were being truncated. Presets is now 16 attributes plus a micronutrients relationship, but the backup payload still carried four (dish/carbs/fat/protein). So a restore wasn't just failing to bring things back — it was handing back a stripped preset, losing fibre, sugars, glycemic index, portion/serving size, tags, image and micronutrients. Anyone who has built up presets through the AI food search would have silently lost all of that enrichment on a restore. Fixed in both directions; micronutrient definitions are looked up by name on the restoring device and created if unknown.
- Override presets lost their Auto ISF block. Auto_ISF is a separate entity keyed by the preset's id, so it never travelled with the preset — a restored preset came back with its per-preset Auto ISF config gone. That one's dosing-relevant.

Everything added to the wire format is Optional, since Swift's synthesized Decodable throws on a missing key rather than using a property default — that's what keeps backups written by older builds decoding, and a nil field leaves the setting alone rather than zeroing it.

One thing stated plainly: the client half of those three additions compiles and runs, but I haven't yet exercised the restore of those specific new fields end-to-end against a backup containing them. The server half is verified. 

Also worth flagging: settings/autoisf.json exists as a constant in Constants.swift but nothing reads or writes it — the global autoISF config genuinely lives in preferences.json and is backed up. Might be worth deleting the constant.

Branch is synced with dev through 77cb19d8d and merges clean.

Jon-b-m:
@scrappy Testing it now. First thing I notice missing - there is no cancel option - no option to skip the onboarding step if you don't need it. 

Skärmavbild 2026-07-30 kl  14 09 36

Jon-b-m:
As an added benefit of the new  Height and weight settings, these we now could use for more individualised macro and micro nutrient RDIs, for instance the recommended dietary daily intake of protein...

scrappy:
Nope, else wouldn't have needed to add it ... its actually stored in

$ grep aiSavePhotosToLibrary *
DatabaseModels.swift:    var aiSavePhotosToLibrary: Bool?

scrappy:
you mean:

./Modules/Main/View/ExistingUserRestoreView.swift: Button("Skip and set 
up fresh") { onDone("") }

Jon-b-m:
> you mean: ./Modules/Main/View/ExistingUserRestoreView.swift: Button("Skip and set up fresh") { onDone("") }
> […](#)
> On 2026-07-30 08:10, Jon B Mårtensson wrote: First thing I notice missing - there is no cancel option - no option to skip the onboarding step if you don't need it.

Not quite, because when selecting this button you still need to go through several additional screens. You currently can't skip this onboarding altogether.

scrappy:
> you mean: ./Modules/Main/View/ExistingUserRestoreView.swift: Button("Skip and set up fresh") { onDone("") }
> […](#)
> On 2026-07-30 08:10, Jon B Mårtensson wrote: First thing I notice missing - there is no cancel option - no option to skip the onboarding step if you don't need it.

Good catch — the skip I pointed you to only exists inside the Existing User flow, so the Welcome screen itself had no way out for someone who doesn't need onboarding at all. Added a "Skip setup — configure everything myself" link under the two cards: it flips the same completion flags as finishing the flow and drops you straight into the app. It deliberately doesn't touch the old firstRun import screen, and the "Skip and set up fresh" option inside Existing User stays, so changing your mind after tapping in still has an exit. Pushed with this update.

PR #1971, opened 2026-07-24
fix: sync Omnipod pod expiry to home header on state updates

degola:
## Summary
- Home could show **No Pod** while Omnipod pump settings still showed a live pod (reservoir / expire time) after the 8.3.0 OmnipodKit migration.
- OmnipodKit calls `pumpManagerDidUpdateState` on every `podState` change, but only emits `didUpdate status` when `PumpManagerStatus`/highlight changes. Pod `expiresAt` lives on `podState`, so the home header’s `pumpExpiresAtDate` was often left stale/`nil`.
- Refresh pod-derived home fields (`pumpExpiresAtDate`, pod age) from `pumpManagerDidUpdateState` (and share that path via `refreshPodDerivedState`).

## Test plan
- [ ] With an active Omnipod DASH pod, open the app and confirm the home header shows reservoir / remaining time (not **No Pod**).
- [ ] Open pump settings and confirm the same expiry/reservoir as on home.
- [ ] Force a pod status refresh / BLE reconnect and confirm home updates without needing a status-only change.
- [ ] Deactivate / no-pod state still shows **No Pod** on home.

### DIFF
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -1028,8 +1028,11 @@ private extension BaseDeviceDataManager {
 
         if let reservoir = KnownPlugins.pumpReservoir(pumpManager) {
             storage.save(reservoir, as: OpenAPS.Monitor.reservoir)
-            broadcaster.notify(PumpReservoirObserver.self, on: processQueue) {
-                $0.pumpReservoirDidChange(reservoir)
+            // setupPump / didUpdateState can run off processQueue; Broadcaster requires it.
+            processQueue.safeSync {
+                broadcaster.notify(PumpReservoirObserver.self, on: processQueue) {
+                    $0.pumpReservoirDidChange(reservoir)
+                }
             }
         }
     }
--- a/FreeAPS/Sources/APS/DeviceDataManager.swift
+++ b/FreeAPS/Sources/APS/DeviceDataManager.swift
@@ -578,6 +578,10 @@ extension BaseDeviceDataManager: PumpManagerDelegate {
             self.pumpManager = newPumpManager
         }
         pumpName.send(pumpManager.localizedTitle)
+        // OmnipodKit notifies state on every podState change, but only emits
+        // status updates when PumpManagerStatus/highlight changes. Pod expiry
+        // lives on podState, so refresh home-header fields here as well.
+        refreshPodDerivedState(from: pumpManager)
     }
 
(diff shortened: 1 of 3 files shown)


yurique (changes requested):
I'm not sure I follow the part about "dual loading"

> OmnipodKit may be loaded from the plugin bundle while FreeAPS also imports the module.
> Prefer typed access when the cast works; otherwise parse `rawState`.

There is only *one* copy of `OmnipodKit.framework` (and other plugins) in the bundle, there is no dual loading (this is explicitly enforced in the `FreeAPS/Scripts/copy-plugins.sh` build step - same as in the `LoopWorkspace`). If this wasn't true - the whole app would have fallen apart long time ago.

---

Inside `func refreshPodDerivedState`:
```swift
processQueue.safeSync {
    broadcaster.notify(PumpReservoirObserver.self, on: processQueue) {
        $0.pumpReservoirDidChange(reservoir)
    }
}
```

you are doing `processQueue.safeSync`.
This function can be called from the main thread, and you acknowledge this in a comment:
> setupPump / didUpdateState can run off processQueue

The potential problem - anything in one of the listeners (one of those `$0.pumpReservoirDidChange(...)`) that happens to also block on main thread ( `persistentContainer.viewContext` / `performAndWait`) (or via some other path) - this can become a deadlock and the app will crash (will get killed by iOS).

So, at the very least this should be `processQueue.async`, *not* `processQueue.safeSync`.

---

`PumpView.body` now calls `KnownPlugins.pumpExpirationDate` on every render (in the situation when there is no pod).

This can force the pump manager do the full serialization of it's state into a dictionary, on the main thread, on every render (while showing "No Pod").

---

These issues are already fixed in the `alpha` branch, but since it is still in (early-ish) testing, we should have this fixed in the `dev` after clean up.

---

My suggestion - 
* keep the changes to bare minimum
* keep the new updating of the POD info in the `pumpManagerDidUpdateState` callback 
* don't change the `setupPump` - this adds additional main thread interactions/issues risks
* don't change the UI code
* drop the unnecessary "double load" workarounds since they are not needed.

yurique:
superseded by https://github.com/Artificial-Pancreas/iAPS/pull/1980

@degola I'll close this PR this time (I don't mean to discourage contributions - but `DeviceDataManager` is a very fragile integration point with LoopKit world, we should be really careful with it).

thomasvargiu:
@yurique I still have the same problem with "No pod" in home page.

PR #1981, opened 2026-07-28
record and display events: device/algorithm failures, etc

yurique:
Screenshots:

image

image

image

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,5 +84,4 @@ ConfigOverride.xcconfig
 branch.txt
 package-lock.json
 node_modules
-oref0
 .temp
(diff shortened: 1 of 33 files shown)


Jon-b-m:
Thanks! Looks good 👍!

Jon-b-m:
I get this build error when testing:
```
Build input file cannot be found: '/Users/jonbm/dev/15juli/iaps/FreeAPS/Sources/Models/oref0/GlucoseEntry0.swift'. Did you forget to declare this file as an output of a script phase or custom build rule which produces it?
```

Jon-b-m:
Could we change the default to hidden? I'm not entirely sure, but I guess most users don't want to always see this in main chart?

yurique:
> Could we change the default to hidden? I'm not entirely sure, but I guess most users don't want to always see this in main chart?

Of course!
(I was sure I made it hidden by default)

PR #1987, opened 2026-08-04
fix meal deletion

yurique:
We are deleting meals in a JSON file, and the dates have been truncated to milliseconds by this time.
The in-memory dates are not truncated to milliseconds, so the delete could not match them.

Edit:`isSameInstant` change was reverted.
~Fix (short-term): compare dates with `date.isSameInstant(otherDate)` (difference is < 1 millisecond).~

Fix (short-term): when carbs and pump events are created - truncate the dates to seconds (truncating to milliseconds was also unreliable - JSON encoding can still change the values).

Also fixing a double-save of meals in BolusStateModel and reverted `Sex.savedSettings` to be a static function (the build was not compiling because the static function was referenced in some file).

### DIFF
--- a/FreeAPS/Sources/Modules/Sharing/SharingStateModel.swift
+++ b/FreeAPS/Sources/Modules/Sharing/SharingStateModel.swift
@@ -17,7 +17,7 @@ extension Sharing {
             subscribeSetting(\.birthDate, on: $birthDate) { self.birthDate = $0 }
             subscribeSetting(\.sexSetting, on: $sexSetting) { self.sexSetting = $0 }
 
-            sex = sex.savedSettings(sexSetting)
+            sex = Sex.savedSettings(sexSetting)
 
             identifier = userToken.getIdentifier()
         }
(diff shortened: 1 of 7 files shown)


FreeAPS/Sources/Modules/Sharing/SharingStateModel.swift
```diff
@@ -17,7 +17,7 @@ extension Sharing {
             subscribeSetting(\.birthDate, on: $birthDate) { self.birthDate = $0 }
             subscribeSetting(\.sexSetting, on: $sexSetting) { self.sexSetting = $0 }
 
-            sex = sex.savedSettings(sexSetting)
+            sex = Sex.savedSettings(sexSetting)
```
Jon-b-m, line 21: The mutable sex is already defined. Without above change to 'Sex' you will not need the 'static' update...

yurique: The build was failing on this line: https://github.com/Artificial-Pancreas/iAPS/blob/alpha/FreeAPS/Sources/Modules/Home/HomeStateModel.swift#L722

So I figured I'd revert that function to be static again, since it fits better (it is a "constructor" function, and doesn't depend on `self` -> perfect fit for `static`)

PR #1998, opened 2026-08-11
libre3 cgm

yurique:
Adding an ⚠️experimental⚠️ support for Libre 3 sensors.

Before trying to use it, please read https://github.com/LoopKit/LibreLoop 

In particular, the caveats about ⚠️alerts⚠️ - same as with Libre 2, the only source of low/high alerts will be iAPS (you won't have critical alerts, etc).

Another note:
* No sensor warm-up countdown during a switch-receiver pairing (duration is sensor-reported for fresh activations)
---

Adding Libre 3 turned out to be challenging - because it is built only for `LoopKit`/next-dev + a `LoopAlgorithm` dependency, which is not compatible with LoopKit/dev which we build against (unlike other CGM/pump managers, which maintain both dev and next-dev branches).

Fortunately, the scope of changes needed to make it work with LoopKit/dev was relatively small.

### Untested as of now

This has not been tested yet. It runs on the simulator and shows the first steps of onboarding, but no actual NFS/BLE, of course. I don't have a real device right now to test it fully.

Kaf from Discord volunteered to test it. 

### LoopAlgorithm shim

First of all, I added a shim `LoopAlgorithm` framework (in `Shims/LoopAlgorithm` directory) - this is needed because there is a lot of `import LoopAlgorithm` in the Libre 3 code and I wanted to keep the diff as small as possible.

### Type aliases

Second, instead of HealthKit's `HKQuantity` and `HKUnit`, next-dev introduces `LoopQuantity` and `LoopUnit` structs which are used by the next-dev modules.

Those structs mirror HealthKit's types almost identically, so the solution was to add the type aliases to the `LoopAlgorithm` shim:
```swift
public typealias LoopQuantity = HKQuantity
public typealias LoopUnit = HKUnit
```

plus a few extension methods on them.

### LoopKit protocols

LoopKit protocols changed to async/await in next-dev. In `LibreLoop` I changed a couple of signatures and callbacks to match the LoopKit/dev protocols:
```swift
// next-dev
public func acknowledgeAlert(alertIdentifier: Alert.AlertIdentifier) async throws {}

// dev
public func acknowledgeAlert(alertIdentifier: LoopKit.Alert.AlertIdentifier, completion: @escaping ((any Error)?) -> Void) {}
```

```swift
// next-dev
         return CGMManagerStatus(hasValidSensorSession: state.sensorSerial != nil && !state.sensorNeedsReplacement,
                                inSensorWarmup: inWarmup,
                                isInoperable: isInoperable,
                                 lastCommunicationDate: state.latestReadingTimestamp,
                                 device: device)

// dev
         return CGMManagerStatus(hasValidSensorSession: state.sensorSerial != nil && !state.sensorNeedsReplacement,
                                 lastCommunicationDate: state.latestReadingTimestamp,
                                 device: device)
```

and removed a bunch of `awaits`, like this:
```swift
// next-dev
Task { await delegate?.issueAlert(alert) }

// dev
Task { delegate?.issueAlert(alert) }
```

### LibreCRKit

`LibreCRKit` was failing to compile with the "compiler unable to type-check" errors - although the code was right. So I rewrote two expressions (same behavior, but xcode is happy).

---

Currently the submodules point to my forks:
* https://github.com/yurique/LibreCRKit/tree/iaps-loop-dev
* https://github.com/yurique/LibreLoop/tree/iaps-loop-dev
(this will need to be copied to `Artificial-Pancreas` org before merge)


Jon-b-m:
@yurique I created a *libre3* branch and forked the two new submodules into Artificial-Pancreas. The new forks now synced with your forks.

PR #2002, opened 2026-08-12
[alpha] Recover from WebContent-process termination/suspension instead of hanging the loop

Jon-b-m:


FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
```diff
@@ -13,6 +14,13 @@ final class WebViewScriptExecutor: Sendable {
         }
     }
 
+    deinit {
+        let view = MainActor.assumeIsolated { self.webView }
```
yurique: `MainActor.assumeIsolated` will crash the process if the current thread if not the main thread, and this  class is not a `@MainActor` in `alpha`.

In a running iAPS - this `deinit` will never happen, because we hold the instance of `WebViewScriptExecutor` forever. It's probably safer to just remove the `deinit` completely.

(If we ever add tests/previews/etc, something that creates and destroys `WebViewScriptExecutor`s dynamically - we'll have to revisit this).

FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
```diff
@@ -31,6 +39,36 @@ final class WebViewScriptExecutor: Sendable {
...
+                window = windowScene.windows.first(where: { $0.isKeyWindow }) ?? windowScene.windows.first
+            }
+        }
+
+        if let window = window {
+            window.addSubview(webView)
+            debug(.openAPS, "WebView successfully attached to window for background protection")
+        } else {
+            // If we still can't find it, we can schedule a retry
+            DispatchQueue.main.async { [weak self] in
```
yurique: This one is a bit scary - if for some reason `window` keeps staying `nil`, this will be a never-ending "loop" on the main thread.

PR #2003, opened 2026-08-13
Watchdog timeout for WebView JS calls — a suspended WebContent process can no longer wedge the loop

scrappy:
While looking into the \"long gaps between Start determine basal and SUGGESTED\" question, I scanned two weeks of logs uploaded to open-iaps.app (83 devices, ~266k loop cycles). Full write-up: https://open-iaps.app/docs/loop-stalls

The biggest failure mode (75% of the 289 wedges found): the app is alive and logging, but the oref pipeline is stuck inside a JS stage. `evaluateFunction` awaits the jsBridge reply with no timeout — when iOS suspends the WKWebView's WebContent process mid-call, the reply never arrives and never errors, so the cycle hangs silently. The two longest wedges (5.2 h and 1.8 h) completed the *moment* the user foregrounded the app, i.e. exactly when iOS resumed the WebContent process.

This PR races the bridge reply against a 90 s watchdog. On timeout, the existing retry path takes over: rebuild the WebView (fresh WebContent process) and retry, so a wedge now costs about a minute instead of hours. The failure log also moves from `print` to `warning(.openAPS, …)` so recoveries are visible in uploaded device logs.

90 s is deliberately generous — healthy calls finish in 1–2 s, and the margin covers autotune-sized work on older hardware (fleet data: only 0.1% of cycles legitimately exceed 60 s).

Related PRs from the same analysis (all independent, all touch this area, happy to rebase whichever merges last): WebContent-terminate recovery, WKUserScript injection, no-loop alert escalation.

### DIFF
--- a/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
+++ b/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
@@ -118,6 +118,13 @@ import WebKit
         }
     }
 
+    /// How long to wait for the JS bridge to reply before declaring the call wedged.
+    /// Healthy calls complete in 1–2 s; the margin covers autotune-sized work on old
+    /// hardware. Without this, a suspended WebContent process leaves the loop hanging
+    /// forever — field logs show silent multi-hour gaps that resolve the moment the
+    /// app is foregrounded (the reply arrives when iOS resumes the WebContent process).
+    private let watchdogTimeout: TimeInterval = 90
+
     private func evaluateFunction(body: String, attempts: Int = 0) async throws -> RawJSON {
         let maxAttempts = 2
         let requestId = UUID().uuidString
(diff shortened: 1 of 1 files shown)


yurique:
@scrappy we have two fix branches:
* https://github.com/Artificial-Pancreas/iAPS/commits/test_protection/ - for `dev` branch (no PR yet, needs testing)
* https://github.com/Artificial-Pancreas/iAPS/pull/2002 - for `alpha` branch

scrappy:
Took a look at both. Your `withTimeout` commit on `test_protection` (62d1122) covers the same ground as this PR — and does a couple of things better (function name in the timeout error, retry cap on the attach loop). 30 s vs my 90 s is a judgment call; determine-basal normally finishes in under a second, so either is generous. I have no attachment to my version — happy to close this one in favor of `test_protection` once it's been tested, this was meant as guidance.

Jon's window-attach commit is interesting because it goes after the root cause (visibility-based WebContent suspension) rather than the symptom. Our fleet data is consistent with that being the right target for the *majority* of wedges — most completed within seconds of `APPLICATION PHASE: active`, i.e. classic suspend/resume.

One caution before treating `test_protection` as the complete fix: there's a minority class the attach trick likely won't cover — wedges that stayed dead *across* foreground transitions (28 of them, 11 devices, in the two-week window), which look like the WebContent process being terminated rather than suspended. Details in my comment on #2004. For those, a timeout alone recovers the loop but the recovery path also needs the script re-injection to be robust — which is what #2005 addresses (the `test_protection` branch still injects bundles via one-shot `evaluateJavaScript`). So I'd see #2004/#2005 as complementary to `test_protection`, not superseded by it.

scrappy:

PR #2004, opened 2026-08-13
Recover from WebContent-process termination instead of hanging the loop

scrappy:
Companion to #2003, from the same fleet-log analysis (https://open-iaps.app/docs/loop-stalls).

The script executor sets no navigation delegate, so when iOS kills the WebView's WebContent process (jetsam under memory pressure — the wedge rate in our data tracks hardware age, iPhone 11/13-era phones being the worst), nothing notices: in-flight JS calls wait forever for a reply that can't come, and the injected oref bundles are gone.

This PR implements `webViewWebContentProcessDidTerminate`: log a warning, fail every pending call immediately (their existing retry path rebuilds the WebView and retries), and replace the WebView with a fresh one.

Independent of #2003 but they complement each other: this handles the case where WebKit *tells* us the process died; the watchdog handles the case where it's silently suspended and nothing is ever delivered.

### DIFF
--- a/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
+++ b/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
@@ -43,6 +43,7 @@ import WebKit
         config.userContentController = contentController
 
         let webView = WKWebView(frame: .zero, configuration: config)
+        webView.navigationDelegate = self
 
         injectConsoleLogHandler(webView: webView)
         loadScripts(webView: webView)
@@ -194,5 +195,29 @@ import WebKit
                 ))
             }
         }
+    }
+}
+
+extension WebViewScriptExecutor: WKNavigationDelegate {
+    /// The WebContent (JS) process died — jetsam under memory pressure, a crash, or iOS
+    /// reclaiming it in the background. Every in-flight call would otherwise wait forever
+    /// for a reply that can no longer arrive, and the injected oref bundles are gone, so
+    /// fail all pending calls now (their retry path rebuilds and retries) and replace the
+    /// WebView with a fresh one.
+    func webViewWebContentProcessDidTerminate(_: WKWebView) {
+        let pending = continuationStreams
+        continuationStreams.removeAll()
+        warning(
+            .openAPS,
+            "WebContent process terminated by iOS — failing \(pending.count) in-flight JS call(s) and rebuilding the WebView"
+        )
+        for (_, continuation) in pending {
+            continuation.finish(throwing: NSError(
+                domain: "WebViewScriptExecutor",
+                code: 510,
+                userInfo: [NSLocalizedDescriptionKey: "WebContent process terminated"]
+            ))
+        }
+        webView = createWebView()
     }
 }


yurique:
This looks interesting, too. But the same question - does iOS ever kill the WebContent process?

Jon-b-m:
> This looks interesting, too. But the same question - does iOS ever kill the WebContent process?

Depends what question you’re asking: 1: can iOS kill a process - yes, 2: is iOS killing iAPS WebContent - this I don’t know yet, but we can’t rule it out.

yurique:
No, sure - iOS can kill *our* process.

I was wondering if there are any traces in the logs of `WebContent` process getting killed while iAPS is still running .

I'm also wondering if `webViewWebContentProcessDidTerminate` is reliable (will it get called)?

Jon-b-m:
> I was wondering if there are any traces in the logs of `WebContent` process getting killed while iAPS is still running.

I have't found any on any of my devices yet...

scrappy:
> I was wondering if there are any traces in the logs of `WebContent` process getting killed while iAPS is still running.

Two things about why you wouldn't find traces even if it's happening:

1. Nothing in the current code writes a log line when WebContent dies — there's no `webViewWebContentProcessDidTerminate` handler, which is exactly the gap this PR fills. iOS records the kill in the *system* log / jetsam reports (visible in Console.app or a sysdiagnose), not in iAPS's own log files. So "no traces on my devices" doesn't rule it out.
2. It also wouldn't show up as failed loops or error reports. If WebContent dies mid-call, `evaluateJavaScript` has already returned successfully; the code is awaiting a `jsBridge` reply that will simply never arrive. No error, no log line — the only visible symptom is a loop cycle that silently never finishes.

So I went back to the fleet scan (83 pod devices, 968 device-days, ~266k loop cycles) and looked specifically at the wedges that **never** completed:

- Of ~289 wedges (>120 s between `Start determine basal` and `SUGGESTED`), ~70 never produced a `SUGGESTED` at all — they ended only with an app relaunch or ran silent to the end of the log.
- 60 of those 70 had the app demonstrably alive the whole time (pump comms, CGM, UI all still logging).
- **28 of those (across 11 devices) spanned at least one `APPLICATION PHASE: active` transition and still didn't recover.** One device went through 10 foreground transitions over 101 minutes while wedged; another stayed wedged 162 minutes across 2 foregroundings.

That last group is the interesting one. Plain suspension recovers on foregrounding — that's exactly what the majority of wedges did (many completed within seconds of `APPLICATION PHASE: active`). Staying dead across repeated foregroundings is what a terminated WebContent process looks like from the app side: the JS context and the pending reply are gone, and nothing short of recreating the webview brings them back. I can't distinguish "killed by iOS" from "JS wedged unrecoverably" with app-side logs alone, but neither recovers without the recreate this PR does — and the hardware pattern (iPhone 11 wedges at ~30× the rate of current-gen phones) fits memory-pressure kills.

> I'm also wondering if `webViewWebContentProcessDidTerminate` is reliable (will it get called)?

Honest answer: I can't confirm from first-hand testing (as I said, these PRs are untested). The documented contract is that it fires when WebKit detects the termination, and if the app itself is suspended at the time, delivery comes when the app resumes. If there are edge cases where it never fires, the timeout (#2003 / your `test_protection` version) is the backstop layer — that's why I split them.

One side benefit of merging this regardless: it makes your original question answerable with data. Once "WebContent process terminated" lands in the device logs, the next fleet scan can count actual terminations instead of us inferring them.

scrappy:

PR #2005, opened 2026-08-13
Inject oref bundles as WKUserScripts so the JS environment survives WebContent relaunches

scrappy:
Third piece of the loop-stall work (#2003, #2004; analysis: https://open-iaps.app/docs/loop-stalls).

The oref bundles and the console hook are injected with one-shot `evaluateJavaScript` calls when the WebView is created. That state lives only in the current WebContent process — after iOS kills and relaunches it, every `freeaps_*` function is undefined and the only recovery is rebuilding the whole executor.

This PR registers everything as `WKUserScript`s on the `WKUserContentController` and loads a blank document to trigger injection. WebKit re-applies user scripts on every document load — including after a WebContent-process relaunch — so the JS environment heals itself. A `__ready__` marker script posted at documentEnd gates calls on the (re)load finishing, so nothing races the injection (previously the first loop after launch raced the bundle injection too; this closes that hole as a side effect).

Touches the same file as #2003/#2004 — merge in any order, I'll rebase the stragglers.

### DIFF
--- a/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
+++ b/FreeAPS/Sources/APS/OpenAPS/WebViewScriptExecutor.swift
@@ -33,24 +33,48 @@ import WebKit
         webView = createWebView()
     }
 
+    /// True once the current WebView has loaded its document and every user script
+    /// (console hook + oref bundles) has been injected. Calls wait on this instead of
+    /// racing the initial page load.
+    private var isReady = false
+    private var readyWaiters: [CheckedContinuation<Void, Never>] = []
+
     private func createWebView() -> WKWebView {
         let contentController = WKUserContentController()
         contentController.add(self, name: "consoleLog")
         contentController.add(self, name: "jsBridge")
         contentController.add(self, name: "scriptError")
 
+        // Inject everything as WKUserScripts rather than one-shot evaluateJavaScript
+        // calls: user scripts are re-applied by WebKit on every document load, so the
+        // oref bundles survive a WebContent-process relaunch instead of vanishing.
+        for source in userScriptSources() {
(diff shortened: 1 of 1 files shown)


yurique:
This looks like a right thing to do.

Do we know if iOS ever kills the WebContent process? (we'd have a lot of failed loops and reports if it did, right?)

scrappy:
> Do we know if iOS ever kills the WebContent process? (we'd have a lot of failed loops and reports if it did, right?)

That's the trap — with the current code a WebContent death produces *no* failure report at all. `evaluateJavaScript` has already returned; the code is awaiting a `jsBridge` reply that will never come. No error, nothing logged — just a loop cycle that silently never finishes. So the absence of reports doesn't tell us anything.

I posted the fleet data on this over in #2004: of the wedges that never completed, 28 (across 11 devices) stayed wedged through one or more foreground-active transitions — up to 10 foregroundings over 101 minutes. Suspension recovers on foregrounding (that's what the majority of wedges did); staying dead across repeated foregroundings is consistent with the WebContent process having been terminated. Not proof — I can't see jetsam from app-side logs — but nothing else in the logs explains that group.

Worth noting this PR pays for itself even if kills turn out to be rare: today the bundles are injected via one-shot `evaluateJavaScript` calls at init with nothing awaiting their completion, so there's also a cold-launch race the `__ready__` gate closes. Same caveat as the others though — untested, treat as guidance.

yurique:
@scrappy I ported your changes from this and other PRs into the existing PR: https://github.com/Artificial-Pancreas/iAPS/pull/2007

I tested it on the phone, and everything works as expected: with the "user-scripts" change + the "empty root html" change - the `webViewWebContentProcessDidTerminate` does get called and iAPS recovers .

scrappy:

PR #2006, opened 2026-08-13
Keep the 'iAPS not active' alert firing until looping resumes, and make it Time Sensitive

scrappy:
Last piece of the loop-stall work (#2003, #2004, #2005; analysis: https://open-iaps.app/docs/loop-stalls).

The missing-loop dead-man's-switch works, but it arms exactly two notifications (20 and 40 min) and then goes quiet — a loop that stays wedged never re-arms the chain, so a multi-hour gap means two banner pings at the start and silence for the rest. Our fleet data has a 5.4 h wedge overnight; two quiet banners at 23:20 and 23:40 are easy to sleep through, and without an interruption level a Sleep Focus swallows them entirely.

Two changes, reusing the existing localized strings (no new keys for Crowdin):

- a third, repeating notification every 60 min, cancelled/re-armed by each loop like the existing two, so the alerts never stop before the looping resumes;
- all three marked `.timeSensitive` (+ the matching entitlement) so they break through Focus modes — a stopped loop is the textbook case for it.

**Review note:** the entitlements addition needs the Time Sensitive Notifications capability on the App ID. Xcode automatic signing handles that by itself; if the fastlane pipeline manages the App ID explicitly it may need the capability enabled there. If that causes friction for the community build, I'm happy to drop the entitlement hunk — the repeating third notification is worth having on its own.

### DIFF
--- a/FreeAPS/Resources/FreeAPS.entitlements
+++ b/FreeAPS/Resources/FreeAPS.entitlements
@@ -12,6 +12,8 @@
 	<array>
 		<string>TAG</string>
 	</array>
+	<key>com.apple.developer.usernotifications.time-sensitive</key>
+	<true/>
 	<key>com.apple.security.application-groups</key>
 	<array>
 		<string>$(APP_GROUP_ID)</string>
--- a/FreeAPS/Sources/Services/UserNotifiactions/UserNotificationsManager.swift
+++ b/FreeAPS/Sources/Services/UserNotifiactions/UserNotificationsManager.swift
@@ -35,6 +35,7 @@ final class BaseUserNotificationsManager: NSObject, UserNotificationsManager, In
         case carbsRequiredNotification = "FreeAPS.carbsRequiredNotification"
         case noLoopFirstNotification = "FreeAPS.noLoopFirstNotification"
         case noLoopSecondNotification = "FreeAPS.noLoopSecondNotification"
+        case noLoopRepeatingNotification = "FreeAPS.noLoopRepeatingNotification"
         case bolusFailedNotification = "FreeAPS.bolusFailedNotification"
         case pumpNotification = "FreeAPS.pumpNotification"
     }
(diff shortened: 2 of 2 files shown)


yurique:
time-sensitive notifications is something I was planning to implement 👍 

I was thinking we should use time-sensitive notifications for all alerts (or most of them at least - hi/low glucose, etc), not just "no loops".

One thing to consider: the users will need to update their identifiers in the Developer account and enable the Time-Sensitive notifications checkbox:
image


Unless there is a way to automate this, we will probably get a storm of failed-build reports.

scrappy:
Agreed on doing time-sensitive across all alerts (hi/low glucose especially) — I kept this PR deliberately minimal, but it'd slot into that.

On the identifier problem: it's mostly automatable. The Fastfile's `identifiers` lane already creates capabilities on the App IDs via Spaceship (`bundle_id.create_capability`), and Spaceship has the constant for this one — `Spaceship::ConnectAPI::BundleIdCapability::Type::USERNOTIFICATIONS_TIMESENSITIVE`. So it's a one-line addition to the FreeAPS entry in that lane. The catch is exactly the one you're pointing at: it only helps users who re-run **Add Identifiers** (and then **Create Certificates** to refresh the profile). Browser-build users who just hit **Build iAPS** on the new tag would fail at signing — so yes, shipped naively this generates the failed-build storm.

Suggested de-risking — split this PR in two:

1. **The repeating reminder needs no entitlement at all.** The hourly `UNTimeIntervalNotificationTrigger` repeat is plain UserNotifications API. That part fixes the actual hole (today the alert goes permanently silent after 2 banners / 40 min) and can merge safely now. My reading of the docs is that `interruptionLevel = .timeSensitive` without the entitlement is silently downgraded to a normal notification rather than being an error (not device-tested — these PRs are untested guidance), so those lines could ride along harmlessly or be dropped, whichever you prefer.
2. **The entitlement + Fastfile capability + docs note** land later as a coordinated piece with your broader time-sensitive-alerts work, in a release whose notes say "re-run Add Identifiers and Create Certificates".

Happy to strip the entitlements hunk from this PR to make it shape 1 — say the word.

PR #2007, opened 2026-08-13
[dev] Recover from WebContent-process termination/suspension instead of hanging the loop

Jon-b-m:
iOS can suspend/kill the WebContent process while iAPS app is in the background.

This PR adds 4 safeguards:
* a timeout for each JS invocation (the loops won't stay stuck forever if WebView is not running)
* attach WebView to the window (this does not prevent the `WebContent` process from being suspended, but hopefully makes it less likely/frequent)
* inject JS-bundles as user-scripts - this makes `WebKit` reinject them after relaunch
* detect and recover from `WebContent` process termination 

### Tested on device
  - timeouts ✅  (3 retries, timing out after 30s, then `Loop failed`)
  - `WebContent` termination: the WebView is recreated and the loop is failed (within ~1s).

### Caveat
- The iAPS process can still get suspended if WebView is suspended (background tasks will most likely expire before the timeouts happen)
- having a real pump with more BLE traffic will mitigate this
- in this case the timeouts will trigger on next BLE activity - the mechanisms will still work, but with a longer delay (5 min vs 1 min)

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -307,6 +307,7 @@
 		38E98A2725F52C9300C0CED0 /* CollectionIssueReporter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2025F52C9300C0CED0 /* CollectionIssueReporter.swift */; };
 		38E98A2925F52C9300C0CED0 /* Error+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2225F52C9300C0CED0 /* Error+Extensions.swift */; };
 		38E98A2D25F52DC400C0CED0 /* NSLocking+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2C25F52DC400C0CED0 /* NSLocking+Extensions.swift */; };
+		F2A11C1230200B0100EF9230 /* Timeout.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2A11C1130200B0100EF9230 /* Timeout.swift */; };
 		38E98A3025F52FF700C0CED0 /* Config.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2F25F52FF700C0CED0 /* Config.swift */; };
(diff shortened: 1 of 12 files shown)


scrappy:
Nice — this is a better-integrated version of the same four safeguards, and the device test answers the one question I couldn't (that `webViewWebContentProcessDidTerminate` really fires once there's a document load behind the user scripts). The `webViewGeneration` guard and the `didFailProvisionalNavigation` logging are both improvements over my versions. I'm closing #2003/#2004/#2005 in favor of this PR, as promised.

Two things I spotted while reading the combined result, one of them worth fixing before merge:

**1. Timed-out calls leak a permanently-suspended task (and the script payload with it).**

`withTimeout` resumes the caller on timeout but never cancels the losing racer, so the inner `Task { ... for try await value in stream ... }` keeps waiting. The catch in `evaluateFunction` then does:

```swift
continuationStreams.removeValue(forKey: requestId)
```

— which removes the continuation *without finishing it*. After that, nothing can ever finish that stream: a late `jsBridge` reply finds no dict entry and is dropped, and the `didTerminate` path only fails continuations still in the dict. Result: every timed-out attempt strands one suspended task holding the stream plus the captured `script` string — and that string contains the full oref invocation payload (glucose history, profile, etc.), so it's not small.

The loop still recovers fine; the problem is accumulation. From the fleet data, the devices that hit 30 s cycles routinely are the old/memory-pressured ones — the same devices jetsam is already squeezing — so a leak per timeout is self-defeating there. One-line fix:

```swift
continuationStreams.removeValue(forKey: requestId)?.finish(throwing: error)
```

Finishing the stream wakes the orphaned task, which throws into the already-decided race (harmlessly dropped) and exits, releasing everything.

**2. Minor: a stale `__iaps_ready__` from a replaced WebView can mark the new one ready early.**

If a WebView is replaced while its document-end user script is in flight, the late ready message sets `isReady = true` for the *new* generation before its scripts have run. It self-heals (the next call errors on an undefined function and retries), so this is a nit — but `WKScriptMessage` carries the sending view, so it's cheap to guard:

```swift
guard message.webView === webView else { return }
```

(Also a garbled doc comment on `replaceWebView`: "Anything waiting on the is failed rather.")

Same standing caveat: I'm reading, not running — you're the one with it on a phone.

Jon-b-m:
Tested in simulator and on real iOS device. 
Limited testing due to needed hotfix.

yurique:
@scrappy thanks! These findings are addressed in a follow-up PR: https://github.com/Artificial-Pancreas/iAPS/pull/2008

PR #2017, opened 2026-08-15
Report memory / process-pressure telemetry with the statistics uploads

scrappy:
Follow-on from the loop-stall work (https://open-iaps.app/docs/loop-stalls and the #2003–#2007 discussion). Two questions came out of that investigation that log files fundamentally can't answer:

1. **Does iOS actually kill the WebContent process?** (@yurique's question on #2004/#2005) — the death is silent, so absence of log traces proves nothing either way. The best I could do was infer it from 28 wedges that survived repeated foregroundings.
2. **Why do older phones wedge at up to ~30× the rate of current ones?** Memory pressure is the suspect, but nothing in our logs measures it.

This PR adds a small `Memory` block to the statistics upload the app already sends (~2–3×/day), so those questions become measurable across the whole fleet instead of inferred from the ~10% of devices that share logs:

* **MetricKit** (`MXAppExitMetric` / `MXMemoryMetric`): OS-recorded counts of jetsam kills, background memory-pressure kills, and watchdog kills, plus daily peak footprint — iOS telling us directly why it terminated the app, no inference.
* **Live gauges** at upload time: current `phys_footprint`, and (full-statistics tier only) device RAM and `os_proc_available_memory` headroom.
* **Two executor counters**: WebContent-process terminations (incremented in the `didTerminate` handler from #2007) and JS watchdog timeouts. These matter doubly now: since #2007 caps wedges at 30 s, the multi-hour log signature I scanned for is gone — these counters are how we keep measuring wedge frequency after the fix hides it, and they measure it from *every* stats device, not just log sharers.

### What each tier sends

Devices on the minimal sharing level today send only `{id, Branch, created_at, Build_Version}`. With this PR the minimal ping gains app-behaviour numbers only — nothing about the person or the phone:

```json
"Memory": {
  "footprint_mb": 69,
  "peak_mb": 201,
  "counters": {
    "8.3.5(276)": {
      "peak_mb": 201,
      "pressure_warnings": 0, "jetsam_exits": 0, "pressure_exits": 0,
      "watchdog_exits": 0, "webcontent_terminations": 0, "js_timeouts": 0
    }
  }
}
```

The full-statistics tier additionally gets `"ram_mb": 7645, "available_mb": 3306` — those two narrow down the phone model class, so they are deliberately excluded from the minimal tier, where the model is not shared. Full field-by-field documentation (source, meaning, tier) is published at **https://open-iaps.app/statistics/memory-diagnostics**, which is also where the collected fleet data is reported.

### Design notes

* **Counters are bucketed per `version(build)`** so events never smear across an upgrade — "did this release make memory pressure better or worse" stays answerable per release. Composite keys because build number alone collides for local Xcode builds (CFBundleVersion stays 1) and version alone can't separate two builds of one release. Newest 6 buckets are retained; old ones freeze and have already been uploaded.
* **MetricKit counts are filed by the payload's own build metadata**, not the running build — MetricKit delivers a day late, so without this, every upgrade would dump the old build's last day of kills onto the new build.
* **Counters are monotonic** (UserDefaults, survives updates, resets only on reinstall); the server diffs successive uploads, so the app carries no upload-state bookkeeping and a failed upload loses nothing.
* MetricKit only delivers on physical devices — on simulators those fields simply stay `nil`. No entitlements, no new permissions, no user-facing strings.

### Tested

Running on my phone (built via Actions) and verified end-to-end: the block arrives in the open-iaps statistics intake and feeds the page above. My own numbers — 69 MB footprint, 201 MB daily peak against a ~3.4 GB allowance, all counters zero — are exactly what a healthy current-gen phone should report; the signal this exists for lives on the older, smaller-RAM devices once the fleet picks it up.

Files: `MemoryMetricsService.swift` (new, Helpers), `Statistics.swift` / `BareMinimum.swift` (optional `Memory` field), `APSManager.swift` (attach to both payloads), `WebViewScriptExecutor.swift` (two one-line counter hooks), `FreeAPSApp.swift` (start the MetricKit subscriber), pbxproj registration.

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -308,6 +308,7 @@
 		38E98A2925F52C9300C0CED0 /* Error+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2225F52C9300C0CED0 /* Error+Extensions.swift */; };
 		38E98A2D25F52DC400C0CED0 /* NSLocking+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2C25F52DC400C0CED0 /* NSLocking+Extensions.swift */; };
 		F2A11C1230200B0100EF9230 /* Timeout.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2A11C1130200B0100EF9230 /* Timeout.swift */; };
+		F2B33E0230220D0300EF9402 /* MemoryMetricsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2B33E0130220D0300EF9401 /* MemoryMetricsService.swift */; };
 		38E98A3025F52FF700C0CED0 /* Config.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2F25F52FF700C0CED0 /* Config.swift */; };
 		38E98A3725F5509500C0CED0 /* String+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A3625F5509500C0CED0 /* String+Extensions.swift */; };
 		38EA05DA261F6E7C0064E39B /* SimpleLogReporter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38EA05D9261F6E7C0064E39B /* SimpleLogReporter.swift */; };
@@ -1005,6 +1006,7 @@
 		38E98A2225F52C9300C0CED0 /* Error+Extensions.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "Error+Extensions.swift"; sourceTree = "<group>"; };
 		38E98A2C25F52DC400C0CED0 /* NSLocking+Extensions.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "NSLocking+Extensions.swift"; sourceTree = "<group>"; };
 		F2A11C1130200B0100EF9230 /* Timeout.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Timeout.swift; sourceTree = "<group>"; };
+		F2B33E0130220D0300EF9401 /* MemoryMetricsService.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MemoryMetricsService.swift; sourceTree = "<group>"; };
 		38E98A2F25F52FF700C0CED0 /* Config.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Config.swift; sourceTree = "<group>"; };
 		38E98A3625F5509500C0CED0 /* String+Extensions.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "String+Extensions.swift"; sourceTree = "<group>"; };
 		38EA05D9261F6E7C0064E39B /* SimpleLogReporter.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SimpleLogReporter.swift; sourceTree = "<group>"; };
@@ -2230,6 +2232,7 @@
 				38FCF3D525E8FDF40078B0D1 /* MD5.swift */,
 				38E98A2C25F52DC400C0CED0 /* NSLocking+Extensions.swift */,
 				F2A11C1130200B0100EF9230 /* Timeout.swift */,
+				F2B33E0130220D0300EF9401 /* MemoryMetricsService.swift */,
 				38C4D33925E9A1ED00D30B77 /* NSObject+AssociatedValues.swift */,
 				3811DE5725C9D4D500A708ED /* ProgressBar.swift */,
 				3811DE5525C9D4D500A708ED /* Publisher.swift */,
@@ -3792,6 +3795,7 @@
 				23888883D4EA091C88480FF2 /* BolusProvider.swift in Sources */,
 				38E98A2D25F52DC400C0CED0 /* NSLocking+Extensions.swift in Sources */,
 				F2A11C1230200B0100EF9230 /* Timeout.swift in Sources */,
+				F2B33E0230220D0300EF9402 /* MemoryMetricsService.swift in Sources */,
 				19C3C72F2E97AE520000BE4D /* SearchResultsView.swift in Sources */,
 				19C3C7302E97AE520000BE4D /* FoodSearchStateModel.swift in Sources */,
 				19C3C7312E97AE520000BE4D /* BarcodeScannerService.swift in Sources */,
--- a/FreeAPS/Sources/APS/APSManager.swift
+++ b/FreeAPS/Sources/APS/APSManager.swift
@@ -1301,7 +1301,8 @@ final class BaseAPSManager: APSManager, Injectable {
                 ),
                 id: getIdentifier(),
                 dob: settings.birthDate,
-                sex: settings.sexSetting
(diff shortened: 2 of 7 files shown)


Jon-b-m:
@scrappy thanks for PR. It’s quite interesting. 201 MB is a lot for this type of app. It would be useful seeing where this comes from…

scrappy:
Worth being precise about what the 201 MB is, because it's easy to misread: it's not app size and not steady usage — it's MetricKit's `peakMemoryUsage`, the *highest instantaneous* physical footprint the app process hit at any single moment in a 24-hour window. The point-in-time samples from the same device tell the other half of the story: steady state is 60–78 MB. Days since the PR opened peaked at 183, 183 and 149 MB. So the picture is: iAPS idles at ~60–80 MB and briefly spikes to 2–3× that once a day.

Also important: this number **excludes the WebContent process** — the oref JS engine runs in a separate OS process with its own memory accounting (that separateness is exactly why iOS can kill it independently). So this is the Swift side only, not "oref uses 200 MB".

> It would be useful seeing where this comes from…

Agreed, and MetricKit can't tell us — it hands over the daily max with no timestamp or attribution. Candidates I'd look at first, from reading the code:

* `statistics()` fetches **every glucose record ever stored** (`DateFilter.total`) and runs the stats math over the full history, twice a day — that one grows with the age of the install.
* Nightly autotune input assembly: weeks of pump/carb/glucose history get serialized into JSON strings in-app before being handed to the webview.
* Home-screen chart rendering on foregrounding.

Two ways to actually answer it: Instruments/Allocations on a dev build is the definitive tool. For in-field attribution, a cheap follow-up would be logging `phys_footprint` at a few checkpoints (loop start/end, autotune, statistics build) into the existing device log — then the daily peak can be lined up against what the app was doing. Happy to do that as a small follow-up PR if you want it; I deliberately kept this one to collection only.

One more datapoint the telemetry produced since the PR opened, and it's a good one: my phone recorded its **first real WebContent termination** — Aug 15 19:46 local, app backgrounded, **zero in-flight JS calls**, and your #2007 handler logged it and rebuilt the WebView in the same second. No loop impact, `js_timeouts` stayed 0. Cross-checked against the uploaded device log and the counter matches the log line exactly. So: iOS reaps an idle WebContent process even on an 8 GB iPhone under no visible pressure — on the phones this PR is really aimed at, that's happening blind today, and this is how we count it.

That's also my honest answer to "is 201 MB a lot": one device can't say — but this metric collected fleet-wide can. If every phone peaks ~200 regardless of hardware, that's the app's natural working set; if it spreads with device class or grows release-over-release, the per-version buckets will show exactly where it moved.

Jon-b-m:
@scrappy Yes, I am looking at Instruments now. II think I need to record for a longer period on a real phone before making any conclusions. I've only tested it briefly and the iAPS app gets really unresponsive and slow when connected to Instruments, so hard to simulate "normal" behaviour while recording...

Just for info the .total is currently 90 days of data. We never process all data, as this would be too costly/slow. The autotune, even if it's using several files, the total of data is just 24 hours (not weeks) and only run once a day (as you mention.

Jon-b-m:
> One more datapoint the telemetry produced since the PR opened, and it's a good one: my phone recorded its first real WebContent termination — Aug 15 19:46 local, app backgrounded, zero in-flight JS calls, and your https://github.com/Artificial-Pancreas/iAPS/pull/2007 handler logged it and rebuilt the WebView in the same second. No loop impact, js_timeouts stayed 0. Cross-checked against the uploaded device log and the counter matches the log line exactly. So: iOS reaps an idle WebContent process even on an 8 GB iPhone under no visible pressure — on the phones this PR is really aimed at, that's happening blind today, and this is how we count it.

Yes, this seems to work fine now.

scrappy:
You're right on both, and I checked to make sure I don't repeat it: `.total` resolves to a 90-day window (`Configs.swift` — the unbounded case is `.all`, which `statistics()` doesn't touch), and autotune's input is a day's worth, once a day. I had read `.total` as all-time — my mistake. That changes the shape of my hypothesis in a useful way: the twice-daily statistics pass is *bounded* (~26k readings at 90 days × 288/day) and stops growing after 90 days of install age, so it can't be responsible for any long-term creep — at most it's a fixed recurring spike. The daily peaks varying 149–201 MB while the stats pass is deterministic suggests a variable contributor (UI/chart sessions would be my guess) on top of whatever fixed baseline the data passes produce.

On Instruments being too intrusive to capture normal behaviour — that's exactly the gap the checkpoint idea covers. Reading `phys_footprint` is a single `task_info` syscall (microseconds, no observer effect), so logging it at a handful of checkpoints — loop start/end, autotune start/end, statistics build, app foreground/background transitions — runs on a normal phone under completely real conditions, and the daily MetricKit peak can then be bracketed between two known activities from the uploaded log. I'll put that together as a small follow-up PR so you have it as an alternative to wrestling with Instruments — it composes with this PR but doesn't depend on it.

And good to hear the WebContent recovery is confirmed working on your side too — between your test and the counted-and-logged live event above, #2007 looks solid.

Jon-b-m:
@scrappy 
> and stops growing after 90 days of install age, so it can't be responsible for any long-term creep

Actually the saved glucose data keeps growing, but we’re just not using any more than 90 days currently. We don’t have any delete entries when passing 90 days currently. At the first steps of implementing CoreData to iAPS (before it was just using the Disk framework to save on file with a maximum date of 24 hours) I did some research regarding this. I couldn’t find any general consensus this was needed or in the the Apple documentation…

scrappy:
@Jon-b-m fair — I conflated "fetch is bounded" with "store is bounded". The 90-day predicate bounds the *working set*, which is what matters for peaks; the store growing on disk shouldn't cost RAM as long as the date fetches hit the index. Agreed: no long-term creep from that path, no purge needed for memory reasons.

First days of the counters in the wild, in case useful — 11 devices reporting (all dev, since 08-14): https://open-iaps.app/statistics/memory-diagnostics

The two 4 GB phones are the story so far:
- **iPhone 13 mini** (libreTransmitter + DASH): 42 WebContent-process terminations in 20 h, 1 background kill of the app (MetricKit).
- **iPhone SE 3** (G7 + Minimed/RileyLink): 19 terminations, 12 system memory-pressure warnings, **5 background kills of the app**, and a MetricKit daily peak of **437 MB** against an 82 MB steady footprint.

The SE 3 shares logs, so I could line the counters up against the day (they match line-for-line):
- All 19 WebContent terminations were `0 in-flight` — iOS reclaiming the idle helper *between* cycles, at one point once per cycle for 40 minutes. Next cycle rebuilds it; no effect on the loop.
- The app kills happened while backgrounded; each time it was dead under a minute and relaunched by the G7's BLE wake. 288 cycles that day, no gap over 9 min. Worth saying out loud: that only works because there's a BLE peripheral to relaunch it — a setup with no BLE wake source would stay dead. That's the real "loop stopped overnight on an older phone" path, and the counters now measure it directly (`pressure_exits`).
- Same phone the week before on a pre-#2007 build: max cycle durations of 20 min and max intervals of 42 min on several days. The 13 mini went from 47-min max cycles (08-12, pre-#2007) to ~3 min on a post-#2007 build. So the terminations were very likely always there; #2007 turned them from 20–47 min stalls into non-events, and this is the first time we can see how often — a lot more than I'd have guessed.

What I can't tell from this build is what spikes an 82 MB process to 437 MB on a 4 GB phone — that's the footprint-checkpoint logging follow-up, which also records the `branch.txt` commit and a start date in each counter bucket (bucketing by build number turned out useless across devices: the same `dev 81a532e` shows up as five different build numbers). Still testing; PR soon.

Preliminary and small (18 dev devices with data on both sides of #2007, ~80 post device-days): cycles ≥20 min went from ~3% of device-days to zero. I'll post the proper same-device before/after when there's enough post data to mean something.

PR #2027, opened 2026-08-20
Footprint checkpoint logging + commit identity in memory counter buckets

scrappy:
Follow-up to #2017, as promised there — two small pieces:

**1. Footprint checkpoint logging** (`FootprintLog.swift`): one debug-log line with the current `phys_footprint` at the moments that matter — loop start/end, autotune start/end, statistics build start/end, and app phase changes (active/inactive/background). This is for the question Instruments can't reasonably answer on a phone that's looping someone's real life: *what* drives the daily memory peaks. Zero cost when nothing reads it; it's just debug lines in the existing log.

First day of data from my own phone (iPhone 17,1) already gives an answer:

| checkpoint | n | min | avg | max |
|---|---|---|---|---|
| loop start | 94 | 42 MB | 45 MB | 49 MB |
| loop end | 94 | 43 MB | 45 MB | 50 MB |
| autotune start → end | 1 | 44 → 44 MB | | |
| statistics start → end | 1 | 45 → 55 MB | | |
| app active (UI open) | 1 | **96 MB** | | |

The background loop is flat at ~45 MB all day — no growth across 94 cycles, autotune free, statistics +10 MB released afterwards. Opening the UI doubles the footprint on the spot. So the 150–200 MB MetricKit daily peaks we're seeing (437 MB on an SE 3 in the fleet) are foreground/UI-session driven, not loop driven — worth knowing before anyone goes hunting for a loop-side leak.

**2. Counter buckets learn their commit** (follows from what the first fleet days showed on #2017): bucketing by version+build number turned out not to aggregate across devices — the build number is each builder's own Actions run counter, so the same `dev 81a532e` shows up as build 482, 278, 154, 56 and 1 on five different phones. Each bucket now records `branch` (the `branch.txt` "<branch> <sha>" string, i.e. the actual code identity) and `since` (when that build first ran on the device, so counts can become per-day rates). A local Xcode rebuild of the same version at build 1 gets its own bucket instead of merging with the previous commit's; MetricKit payloads (which know version+build but not commit) file into the newest matching bucket. `branch.txt` parsing moved to a shared `Bundle.gitBranch` so APSManager and the bucket code read it the same way.

Both additive; the server side already handles old and new payload shapes. Running on my phone since yesterday — the numbers above and the per-commit fleet view at https://open-iaps.app/statistics/memory-diagnostics are produced by this build.

### DIFF
--- a/FreeAPS.xcodeproj/project.pbxproj
+++ b/FreeAPS.xcodeproj/project.pbxproj
@@ -309,6 +309,7 @@
 		38E98A2D25F52DC400C0CED0 /* NSLocking+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 38E98A2C25F52DC400C0CED0 /* NSLocking+Extensions.swift */; };
 		F2A11C1230200B0100EF9230 /* Timeout.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2A11C1130200B0100EF9230 /* Timeout.swift */; };
 		F2B33E0230220D0300EF9402 /* MemoryMetricsService.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2B33E0130220D0300EF9401 /* MemoryMetricsService.swift */; };
+		F2C44F0230230E0400EF9502 /* FootprintLog.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2C44F0130230E0400EF9501 /* FootprintLog.swift */; };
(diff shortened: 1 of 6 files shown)


scrappy:
@Jon-b-m ... can this get added in before next main release?   it improves the granularity of the memory vs just "what was the highest during the day", by adding a per loop entry to the log to show path through the day ...

Jon-b-m:
> @Jon-b-m ... can this get added in before next main release?   it improves the granularity of the memory vs just "what was the highest during the day", by adding a per loop entry to the log to show path through the day ...
> 
> 

Yes, I think so. I just need to test it for some time first. Thanks!

PR #2028, opened 2026-08-20
Report the raw machine identifier instead of maintaining a device-name table in-app

scrappy:
The identifier→marketing-name switch in `UIDevice+Extensions.swift` stopped at `iPhone16,2`, so 2023-and-older phones report "iPhone 15 Pro Max"-style names while every 16/17-series phone reports a raw `iPhone17,x`/`iPhone18,x` identifier. That half-mapped state means every consumer has to handle both forms — the Pro Max layout checks in `AutoISFHistoryView`/`DynamicHistoryView` were already comparing against a mix of raw identifiers and mapped names, including one (`"iPhone 17 Pro Max"`) the switch never actually produces, so that comparison could never match.

Rather than keep growing the table every hardware cycle, this removes it: `getDeviceId` now returns the raw machine identifier, always. The statistics site maps identifiers to marketing names at display time (already deployed), which also covers every older build that shipped with a stale table — something an in-app table can never do retroactively. The Pro Max screen checks now compare raw identifiers only (`iPhone16,2` / `17,2` / `18,2`).

Net effect: −70 lines, one canonical device string, and new-hardware support stops needing an app change at all.

### DIFF
--- a/FreeAPS/Sources/Helpers/UIDevice+Extensions.swift
+++ b/FreeAPS/Sources/Helpers/UIDevice+Extensions.swift
@@ -10,75 +10,7 @@ extension UIDevice {
             return identifier + String(UnicodeScalar(UInt8(value)))
         }
 
-        func mapToDevice(identifier: String) -> String {
-            switch identifier {
-            case "iPhone10,1":
-                return "iPhone 8"
-            case "iPhone10,2":
-                return "iPhone 8 Plus"
-            case "iPhone10,4":
-                return "iPhone 8"
-            case "iPhone10,5":
-                return "iPhone 8 Plus"
-            case "iPhone10,6":
-                return "iPhone X"
-
-            case "iPhone11,2":
-                return "iPhone Xs"
-            case "iPhone11,8":
-                return "iPhone XR"
-
-            case "iPhone12,1":
-                return "iPhone 11"
-            case "iPhone12,5":
-                return "iPhone 11 Pro Max"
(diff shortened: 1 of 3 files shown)


Jon-b-m:
@thanks. Looks good, but needs testing to see if UI behaves as expected. Have you tested with several devices?

scrappy:
Only have the one physical device (been running this build on it since the 19th, history views look right). But this one can be reasoned through rather than needing a device matrix — the only UI the mapping ever fed is the Pro Max inset checks in the two history views:

- 15 Pro Max (iPhone16,2): old code mapped it to "iPhone 15 Pro Max", which matched the string compare → inset 25. New code matches iPhone16,2 in the list → same values.
- 16 / 17 Pro Max (iPhone17,2 / iPhone18,2): the table was never updated past the 15 Pro Max, so for these getDeviceId already returned the raw identifier and the old compares matched on that → unchanged.
- Every other model: matched neither before nor after → default 15 either way.

That half-stale table is actually what made those checks messy — they had to match a mix of pretty names and raw identifiers, and one compare (`"iPhone 17 Pro Max"`) was dead because nothing could ever produce that string. With raw identifiers everywhere, supporting a future Pro Max is one identifier added to one list.

So layout should be pixel-identical on every model. Happy to leave it open if someone with a 15/16/17 Pro Max wants to eyeball the history-chart alignment to confirm, but there's no code path that changes for them.

PR #2040, opened 2026-08-28
improved CGM calibration

yurique:
This PR improves the CGM calibration screen.

## Older meter readings

Older meter readings can be entered, specifying the time of the reading

Previously a calibration was always paired with the newest CGM reading. Now there is an optional field to specify the time: the raw sensor value is interpolated to that moment, refusing to pair across a gap over 10 minutes, across a sensor session boundary, or before the stored history begins. 

Leaving the time off behaves as before, except it now says so when the newest reading is more than 4.5 minutes old instead of silently pairing against it. 

A warning appears when glucose was moving faster than 1 mg/dL/min around that time, since the sensor's lag then goes straight into the fit.

## Recalibrating the glucose.json

A new button to apply the updated calibration to the current glucose history.

Applies the current slope and intercept to the readings already stored for the current sensor session, in both the glucose file and Core Data.

The `Remove All and reset history` button does the inverse: clears the calibrations and restores those readings to the sensor's raw values. 

## Two new options (toggles)

(both off by default)

* `Robust fit`: takes the slope from the median of the slopes through every pair of calibrations (Theil–Sen - https://en.wikipedia.org/wiki/Theil%E2%80%93Sen_estimator) and the intercept from the median of the residuals, instead of least squares. Useful when one calibration is bad - taken on a ramp, a poor strip, a typo: least squares spreads that error across the whole line, while the median simply outvotes it. (it will not improve anything when the
  calibrations are all good)

* `Allow a wider correction`: appears only when a fit is actually being clamped.

### Other

The slope is no longer estimated when the calibrations span less than 50 mg/dL of sensor value. In this case, an offset is applied instead, with an explanation for the user, because a slope fitted across a narrow range is mostly noise leveraged into a large correction. 

### Other CGMs

Calibrations are now available for other CGMs as well, not just Libre 2/3.

---

Validated against 37 finger sticks with a Libre 3 with 3 good calibration points: raw -> calibrated took the Libre from 23.1% to 5.8% MARD. 

<details>
  <summary>Screenshots 👇 </summary>

Screenshot 2026-08-28 at 22 56 18
Screenshot 2026-08-28 at 22 55 50

</details>


tim2000s:
I wanted to remind everyone that issues with calibration of Libre sensors via open source tools put someone in hospital in the UK in severe hypoglycemia and resulted in the FDA issuing this: https://www.fda.gov/news-events/press-announcements/fda-warns-against-use-unauthorized-devices-diabetes-management

darrelld3:
Libre has also recalled a number of sensors with accuracy problems: https://www.fda.gov/medical-devices/medical-device-recalls-and-early-alerts/glucose-monitor-sensor-recall-abbott-diabetes-care-removes-certain-freestyle-libre-3-and-freestyle

PR #2050, opened 2026-09-16
Swift6 refactoring 1

Jon-b-m:
Remove many Swift6 warnings with many small updates.

There still are some warnings remaining in OpenAPS.swift requiring a bigger refactoring, because its injected service protocols don’t declare Sendable contracts (yet).


yurique:
We have most (if not all) of these warnings fixed in the `alpha` branch, and this will most likely create a lot of merge conflicts.

yurique:
hmm there are a few things here that are going to break things (or even crash the app under the "right" circumstances)

Jon-b-m:
> We have most (if not all) of these warnings fixed in the `alpha` branch, and this will most likely create a lot of merge conflicts.

Yes, we might want to merge alpha first. Do you mind making a PR (alpha to dev?).

Jon-b-m:
> hmm there are a few things here that are going to break things (or even crash the app under the "right" circumstances)

Do you have any examples? I’m running it in vivo since a couple of days…

yurique:
* removing pump or cgm - didSet will happen off the main thread and the precondition in `didSet` will probably crash 
* `beginBackgroundTask` pass `expirationHandler: nil` - if the background task happens to actually expire (although it's unlikely) - the app will be killed 
* `MainActor.assumeIsolated` in the `BaseAPSManager.init` - if that init happens not to be on main - crash (if the app is started from a shortcut, for example)

among other things:

* Stale Live Activities are never cleaned up
* timeouts for the JS function - looks like they will never fire now (because of the task group), but I'd need to look more carefully into that
* DataTable publishes off the main thread 
* formatters became computed vars - will be created over and over again for each date parsing
* the code removed from the app group CGM (xdrip) should not have been removed; it was unused - because it's a bug in dev - but the bug is fixed in alpha, and all that code is actually used (to trigger a "heartbeat" on BT activity, which is just accidentally missing in dev)

--
The list is incomplete. The issues can be fixed one by one, but it looks like the xcode assistant really doesn't know what it's doing :( - just making the warnings "go away" without fixing the underlying issues and/or code/app structure is worthless at best, and more likely - just harmful.

And - this touches 40+ files which are also changed/updated/moved in alpha - resolving the conflicts will be a nightmare.

Jon-b-m:
Closing. New effort will be made when alpha is merged.

